PR #91, opened 2024-02-12
fix: set optionVar value to default before running tests, small additional fixes

rmv:
- Bundle tests rely on a maya option var that is saved in preferences, not scene file.
- When the report file path is shown, the report is not yet created.
- Macos does not show Qt dialog window titles.
- Coverage tests include some test code.

- Set the option Var to its default value before running the tests.
- Before reporting the report file location, make sure it's fully written out.
- Repeat the title info in dialog body.
- Omit test code from coverage testing.

More reliable output bundle tests, minor qol fixes for devs.

Ran the bundle output tests in failure and success modes.

Yes.

No

No


jericht:
> I ended up with a very similar fix in #103 that got merged and didn't see that the work was already done here. ~Closing this.~ actually there's other fixes in here that we want in

@epmog Which other fixes did we want to keep in here?

epmog:
> > I ended up with a very similar fix in #103 that got merged and didn't see that the work was already done here. ~Closing this.~ actually there's other fixes in here that we want in
> 
> @epmog Which other fixes did we want to keep in here?

The changes that aren't yet merged ;) I believe they're QoL improvements around the tests (coverage/displaying results). Feel free to take this PR over.

jericht (approved):

jairaws:
Thanks for your contribution. There are version conflicts that need to be resolved.

leongdl:
1 year old PR with no updates. Please re-open as needed to rebase to altest code.

PR #164, opened 2024-07-10
fix: when listing layers make sure referenced layers are active.

### LINKED ISSUE #160: Bug: Submitter/Adaptor fails to open scene with "... is a layer from a referenced scene and is not allowed to be the current layer"
### Expected Behaviour

Both the submitter and adaptor should not attempt to switch to a render layer that they are not allowed to switch it to.

### Current Behaviour

When launching the submitter, a pop-up appears with a stack trace:

```
Exception caught:
Traceback (most recent call last):
  File "*REDACTED*/deadline/client/ui/__init__.py", line 38, in gui_error_handler
      yield
  File "*REDACTED*/deadline/maya_submitter/mel_commands.py", line 66 in doIt
      DeadlineCloudSubmitterCmd.dialog = show_maya_render_submitter(
  File "*REDACTED*/deadline/maya_submitter/maya_render_submitter.py", line 473, in show_maya_render_submitter
      set_current_render_layer(render_layer_name)
  File "*Redacted*/deadline/maya_submitter/render_layers.py", line 43, in set_current_render_layer
      maya.cmds.editRenderLayerGlobals(currentRenderLayer=render_layer_name
RuntimeError: Maya command error
```
with this in maya console:
```
Warning: '*REDACTED*:defaultRenderLayer' is a layer from a referenced scene and is not allowed to be the current layer.
```

In the adaptor, when it tries to set the current render layer:

```
ADAPTOR_OUTPUT: STDERR: Warning: 'defaultRenderLayer1' is a layer from a referenced scene and is not allowed to be the current layer.
ADAPTOR_OUTPUT: STDERR: Traceback (most recent call last):
ADAPTOR_OUTPUT: STDERR:   File "/tmp/openjd/session-340a6dd6816d4f56ad606fa7e24e672ecbc8nkar/venv/lib/python3.9/site-packages/deadline/maya_adaptor/MayaClient/maya_client.py", line 70, in <module>
ADAPTOR_OUTPUT: STDERR:     main()
ADAPTOR_OUTPUT: STDERR:   File "/tmp/openjd/session-340a6dd6816d4f56ad606fa7e24e672ecbc8nkar/venv/lib/python3.9/site-packages/deadline/maya_adaptor/MayaClient/maya_client.py", line 66, in main
ADAPTOR_OUTPUT: STDERR:     client.poll()
ADAPTOR_OUTPUT: STDERR:   File "/tmp/openjd/session-340a6dd6816d4f56ad606fa7e24e672ecbc8nkar/venv/lib/python3.9/site-packages/openjd/adaptor_runtime_client/client_interface.py", line 181, in poll
ADAPTOR_OUTPUT: STDERR:     self._perform_action(action)
ADAPTOR_OUTPUT: STDERR:   File "/tmp/openjd/session-340a6dd6816d4f56ad606fa7e24e672ecbc8nkar/venv/lib/python3.9/site-packages/openjd/adaptor_runtime_client/client_interface.py", line 202, in _perform_action
ADAPTOR_OUTPUT: STDERR:     action_func(a.args)
ADAPTOR_OUTPUT: STDERR:   File "/tmp/openjd/session-340a6dd6816d4f56ad606fa7e24e672ecbc8nkar/venv/lib/python3.9/site-packages/deadline/maya_adaptor/MayaClient/render_handlers/arnold_handler.py", line 83, in set_render_layer
ADAPTOR_OUTPUT: STDERR:     maya.cmds.editRenderLayerGlobals(currentRenderLayer=render_layer_name)
ADAPTOR_OUTPUT: STDERR: RuntimeError: Maya command error
openjd_fail: Error encountered while starting adaptor: Maya encountered an error and was not able to complete initialization actions.
ERROR: Entrypoint failed: openjd_fail: Error encountered while starting adaptor: Maya encountered an error and was not able to complete initialization actions.
```

The two issues are related but slightly different:

1. adaptor - we're passing in a display name, which Maya _may_ resolve to a non-valid render layer.
2. submitter - tries to get all non-referenced layers, but is unable to filter them out properly.

### Reproduction Steps

I've attached a minimal reproduction case, but I've also included a README that I'll paste here so you can recreate it yourself without using my scene files:

```
Folder layout

1. open_this_scene.mb
2. imported_scene.mb
3. ref.mb

Scene layout:

1 imports 2 without a namespace
2 creates a reference to 3

What this does:

1. There are now 3 renderLayers returned in open_this_scene via maya.cmd.ls(type="renderLayer")
    maya.cmds.ls(type="renderLayer")
    # Result: ['defaultRenderLayer', 'imported_scene_defaultRenderLayer', 'ref:defaultRenderLayer']
2. If you filter out via `isNodeReferenced` you're left with:
    # Result: ['defaultRenderLayer', 'imported_scene_defaultRenderLayer']
3. The display names of these layers (melscript):
    renderLayerDisplayName "defaultRenderLayer"
    // Result: masterLayer
    renderLayerDisplayName "imported_scene_defaultRenderLayer"
    // Result: masterLayer
4. Swapping to 'otherscenewithref_defaultRenderLayer':
    maya.cmds.editRenderLayerGlobals(currentRenderLayer="imported_scene_defaultRenderLayer")
    # Warning: 'imported_scene_defaultRenderLayer' is a layer from a referenced scene and is not allowed to be the current layer.
    # Error: RuntimeError: file <maya console> line 3: Maya command error
```

[referenced_render_layer_issue.zip](https://github.com/user-attachments/files/15876284/referenced_render_layer_issue.zip)

### Code Snippet

Where the submitter grabs all render layers and attempts to switch to all of them

https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/maya_render_submitter.py#L463-L473

Where the adaptor takes a display name and attempts to get the right unique name:

https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py#L57-L73

Here's how you can visualize it without using the submitter/adaptor directly:

Error:

```
'defaultRenderLayer1' is a layer from a referenced scene and is not allowed to be the current layer.
```

```python
import maya
maya.cmds.ls(":*", type="renderLayer")

# output
['defaultRenderLayer', 'defaultRenderLayer1']
```

```python
import maya

def _get_render_layer_display_name(render_layer_name: str) -> str:
    return maya.mel.eval(f'renderLayerDisplayName "{render_layer_name}"')

for name in maya.cmds.ls(":*", type="renderLayer"):
    if not maya.cmds.referenceQuery(name, isNodeReferenced=True):
        print(f"{name} - {_get_render_layer_display_name(name)}")

# output
defaultRenderLayer - masterLayer
defaultRenderLayer1 - masterLayer
```

And then given the display name `masterLayer` Maya will then resolve it (mimicing the adaptor)
```python
# data is "defaultRenderLayer"
render_layer_name = self.get_render_layer_to_render(data)
# render_layer_name is "masterLayer"
if render_layer_name:
    maya.cmds.editRenderLayerGlobals(currentRenderLayer=render_layer_name)
```
with the the referenced one (not sure how Maya determines which one it will resolve to, latest?), which then fails the render or submitter opening with:

```
'defaultRenderLayer1' is a layer from a referenced scene and is not allowed to be the current layer.
```

rmv:
Submitter won't open with Maya files that reference files with inactive additional render layers.

When checking valid render layers make also sure that they are connected to the current layer manager.

No negative impact.

Added a job bundle test, tested with a customer maya scene and with a repro.

Yes, all passed.

No need to, bug fix.

No

----


erico-aws:
Should we add some unit tests for this?

job_bundle_output_tests/referenced_layers/expected_job_bundle/template.yaml
```diff
@@ -0,0 +1,176 @@
...
+    control: CHECK_BOX
+    label: Error on License Failure
+    groupLabel: Arnold Renderer Settings
+  description: Whether to produce an error when there is an Arnold license failure.
+  default: 'false'
+  allowedValues:
+  - 'true'
+  - 'false'
+steps:
+- name: masterLayer
```
lucaseck, line 104: nit: `masterLayer` -> `mainLayer`?

job_bundle_output_tests/referenced_layers/scene/scene_file_to_reference.mb
lucaseck, line 1: Can this be converted to a .ma file instead of binary?

src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
```diff
@@ -57,11 +57,19 @@ def get_camera_to_render(self, data: dict) -> list[str]:
     def get_render_layer_to_render(self, data: dict) -> Optional[str]:
         display_name = data.get("render_layer")
         if display_name:
+            # get render manager
+            filter = maya.cmds.itemFilter(byType="renderLayerManager")
+            render_manager = maya.cmds.lsThroughFilter(filter)[0]
```
lucaseck: Not super familiar with Maya, can there only ever be one renderLayerManager or is [0] always the correct index?

rmv: Good question. Looks like not only is it guaranteed that there will at least be one render layer manager, but it is guaranteed to be called "renderLayerManager". So the code can be simplified.

There can be other nodes of the same type, but they won't be the actual manager.

job_bundle_output_tests/referenced_layers/expected_job_bundle/template.yaml
rmv: Note that this is a default Maya term for the main layer.

job_bundle_output_tests/referenced_layers/scene/scene_file_to_reference.mb
rmv: I kept it as binary as I copied it over from the other tests, but I can do that. In general some DCCs will use binary formats with no human readable option.

rmv:
Pushed some fixes:
- mb to ma: I think we can get rid of those files to reference, work for another PR
- simplified how renderLayerManager node is obtained.

rmv:
@erico-aws was asking:
> Should we add some unit tests for this?

I added a job bundle output test to cover this issue.

I think the code that gets the render layers is not very suited for a unit test, it mainly consists of calls to Maya's API, once we mock up those there would be nothing left to test:

```
# ignore referenced and disconnected layers
render_manager = maya.cmds.ls("renderLayerManager")[0]
render_layer_map = {
    maya.mel.eval(f'renderLayerDisplayName "{render_layer_name}"'): name
    for name in maya.cmds.ls(type="renderLayer")
    if not maya.cmds.referenceQuery(name, isNodeReferenced=True)
    and maya.cmds.listConnections(name, t="renderLayerManager")[0] == render_manager
}
```

job_bundle_output_tests/referenced_layers/scene/scene_file_to_reference.mb
lucaseck: True, in those cases its worth looking into if we can create the scene conditions via the DCCs api to avoid uploading binary files.

lucaseck (approved):

erico-aws (approved):

PR #175, opened 2024-09-19
ci: add bash shell to bump and release actions

moorec-aws:
A shell needs to be specified for the bump and release hook actions

Specify bash for the shell

allows actions to run shell commands

Yes, in a developer github account

Not required, this is a CI change.

No

No

----

### DIFF
--- a/.github/actions/bump_precommit_hook/action.yaml
+++ b/.github/actions/bump_precommit_hook/action.yaml
@@ -8,6 +8,7 @@ runs:
   using: composite
   steps:
   - name: Update plugin Version
+    shell: bash
     run: |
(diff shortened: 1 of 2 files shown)


epmog (approved):

AWS-Samuel (approved):

ddneilson (approved):

.github/actions/bump_precommit_hook/action.yaml
```diff
@@ -8,6 +8,7 @@ runs:
   using: composite
   steps:
   - name: Update plugin Version
+    shell: bash
```
jairaws, line 11: Out of curiosity, what's the behavior without the shell specification? Was the action failing?

PR #176, opened 2024-09-25
feat: Add the ability to specify a render region

mwiebe:
See also the related PR https://github.com/aws-deadline/deadline-cloud-samples/pull/37

Alison Hardy's blog post https://aws.amazon.com/blogs/media/create-a-tile-rendering-job-with-modifications-for-aws-deadline-cloud/
does a great job walking through the development process to create a tile render job. I gave her some ideas for
how to update her deadline-cloud-for-maya change to get it merged, but she didn't have sufficient time at
the end of her internship to finish that.

I've completed the changes I suggested to switch the tiling parameters into more general region render parameters.

This builds on the work of @alichiba. I have adjusted the change from tile index parameters into region render parameters. This works for the tile render case as well as for other region rendering.

Also refactor how the image resolution gets set, switching it from render function parameters to setting properties in the scene.

Commit from @alichiba:

feat: add tile rendering configuration to arnold handler

Once we merge a corresponding change to provide a tile rendering sample job bundle, customers will be able to run tile renders with Maya/Arnold and FFmpeg on Deadline Cloud.

I built my changes into a `maya-openjd` conda package that I published to an S3 conda channel, and submitted a tile render job and the turntable job. Both of these use the adaptor, without using the submitter.

I tried to, but the `hatch` command does not succeed.

```

Timestamp: 2024-09-25T17:28:33.011626-07:00
Running job bundle output test: C:\Dev\deadline-cloud-for-maya\job_bundle_output_tests\cube

cube
Test succeeded

Timestamp: 2024-09-25T17:28:34.822302-07:00
Running job bundle output test: C:\Dev\deadline-cloud-for-maya\job_bundle_output_tests\layers

layers
Test succeeded

Timestamp: 2024-09-25T17:28:36.332334-07:00
Running job bundle output test: C:\Dev\deadline-cloud-for-maya\job_bundle_output_tests\layers_no_variation

layers_no_variation
Test succeeded

Timestamp: 2024-09-25T17:28:37.027234-07:00
Running job bundle output test: C:\Dev\deadline-cloud-for-maya\job_bundle_output_tests\referenced_layers

referenced_layers
Test succeeded

Timestamp: 2024-09-25T17:28:38.239517-07:00
Running job bundle output test: C:\Dev\deadline-cloud-for-maya\job_bundle_output_tests\renderman
Timestamp: 2024-09-25T17:28:40.213306-07:00

```

Updated the schemas for the init-data/run-data additions.

No, this adds new options without removing/changing the existing ones.

----

### DIFF
--- a/src/deadline/maya_adaptor/MayaClient/render_handlers/arnold_handler.py
+++ b/src/deadline/maya_adaptor/MayaClient/render_handlers/arnold_handler.py
@@ -24,7 +24,7 @@ class ArnoldHandler(DefaultMayaHandler):
             data (dict): The data given from the Adaptor. Keys expected: ['frame']
 
         Raises:
-            RuntimeError: If no camera was specified and no renderable camera was found
+            RuntimeError: If no camera was specified and no renderable camera was found. If a render region is partially specified.
         """
         frame = data.get("frame")
         if frame is None:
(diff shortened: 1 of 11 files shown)


src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
```diff
@@ -94,9 +97,31 @@ def start_render(self, data: dict) -> None:
...
+            )
+
+        if self.image_width is not None:
+            maya.cmds.setAttr("defaultResolution.width", self.image_width)
+            print(f"Set image width to {self.image_width}", flush=True)
+        if self.image_height is not None:
+            maya.cmds.setAttr("defaultResolution.height", self.image_height)
+            print(f"Set image height to {self.image_height}", flush=True)
+
+        region = [
```
jairaws, line 116: Q: Why do we retrieve and validate the region values, if we don't use them afterwards.

src/deadline/maya_adaptor/MayaClient/render_handlers/arnold_handler.py
```diff
@@ -33,18 +33,42 @@ def start_render(self, data: dict) -> None:
...
+            print(f"Set imageFilePrefix to {output_file_prefix}", flush=True)
+
+        if self.image_width is not None:
+            maya.cmds.setAttr("defaultResolution.width", self.image_width)
+            print(f"Set image width to {self.image_width}", flush=True)
+        if self.image_height is not None:
+            maya.cmds.setAttr("defaultResolution.height", self.image_height)
+            print(f"Set image height to {self.image_height}", flush=True)
+
+        region = [
```
jairaws, line 51: nit: I feel we would benefit by using a data class to model the region. The current implementation seems error prone. Did we choose this simpler approach because `region` is only used within these few lines of code?

jairaws (dismissed):

src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
mwiebe: Because if the job tries to do a region render when the render handle doesn't support it, the render should fail instead of producing incorrect output.

src/deadline/maya_adaptor/MayaClient/render_handlers/arnold_handler.py
mwiebe: The current implementation is to keep things simple, yes. An advantage of this over a dataclass is that it has better cohesion compared to splitting it across a dataclass definition and where it is used.

jairaws (approved):

src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
```diff
@@ -63,6 +63,29 @@ def start_render(self, data: dict) -> None:
...
+            print(f"Set image height to {self.image_height}", flush=True)
+
+        region = [
+            data.get(field)
+            for field in ("region_min_x", "region_max_x", "region_min_y", "region_max_y")
+        ]
+        if any(v is not None for v in region):
+            raise RuntimeError(
+                "MayaClient: A region render was specified, but region rendering support is not implemented for the selected renderer."
+            )
```
crowecawcaw, line 87: Would it make sense to put this in a common function since it's copied a few times?

crowecawcaw (approved):

src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
mwiebe: Yeah, there's more that could be done to clean up how the implementation is split between the base class and each renderer-specific subclass. I think that will best be done as a separate PR.

PR #190, opened 2024-10-28
fix: frame override checkbox does not enable frame list in Maya 2025

epmog:

Maya 2025 support was just added that uses PySide6. PySide6 changed how they handle enum values: https://doc.qt.io/qtforpython-6/considerations.html#the-new-python-enums. So the way we're checking does not enable the frame list properly.

PR for inspiration: https://github.com/aws-deadline/deadline-cloud-for-blender/pull/90

It works!

```
hatch run lint
hatch run test
```

Will test manually first before merging.

N/A, this is for the UI, not the job bundles.

N/A

Nope

----

### DIFF
--- a/src/deadline/maya_submitter/ui/components/scene_settings_tab.py
+++ b/src/deadline/maya_submitter/ui/components/scene_settings_tab.py
@@ -200,4 +200,4 @@ class SceneSettingsWidget(QWidget):
         """
         Set the activated/deactivated status of the Frame override text box
         """
-        self.frame_override_txt.setEnabled(state == Qt.Checked)
+        self.frame_override_txt.setEnabled(Qt.CheckState(state) == Qt.Checked)


mwiebe (approved):

jericht (approved):

godobyte (approved):

jairaws (approved):

leongdl (approved):

src/deadline/maya_submitter/ui/components/scene_settings_tab.py
```diff
@@ -200,4 +200,4 @@ def activate_frame_override_changed(self, state):
         """
         Set the activated/deactivated status of the Frame override text box
         """
-        self.frame_override_txt.setEnabled(state == Qt.Checked)
+        self.frame_override_txt.setEnabled(Qt.CheckState(state) == Qt.Checked)
```
leongdl, line 203: Super nit, can 199 state use a type so bugs like this can be found via `lint`

leongdl (approved):

PR #192, opened 2024-10-30
fix: maya 2025 submitter failed to load due to qt setMargin being removed

epmog:

Tried submiting jobs from Maya 2025 and ran into a stack trace when loading the scene settings tab. PySide6 was complaining that setMargin no longer exists.

It was deprecated in qt5, and removed in qt6

ref: https://doc.qt.io/qt-5/qlayout-obsolete.html#setMargin

Recommendation is to use setContentsMargin instead, which we already were. So I've deleted the problematic line.

I've also updated the dev install (`hatch run install`) to support Maya 2025 which uses python 3.11. On macOS, 2023 was the last year that relied on rosetta, so I've fixed the dependencies discovery for it.

Maya 2025 can launch and submit jobs!

Admittedly, I did not test the render. 

But I did test the job bundle output tests. Looks like arnold now defaults to true for failing on missing licenses, so I'm now forcibly setting the option in the scenes we have. We probably need to add the ability to differentiate expected output per version.

```

Timestamp: 2024-10-30T15:50:50.623033-05:00
Running job bundle output test: /Users/morgane/dev/github/deadline-cloud-for-maya/job_bundle_output_tests/layers

layers
Test succeeded

Timestamp: 2024-10-30T15:50:51.401506-05:00
Running job bundle output test: /Users/morgane/dev/github/deadline-cloud-for-maya/job_bundle_output_tests/referenced_layers

referenced_layers
Test succeeded

Timestamp: 2024-10-30T15:50:51.746398-05:00
Running job bundle output test: /Users/morgane/dev/github/deadline-cloud-for-maya/job_bundle_output_tests/renderman
Timestamp: 2024-10-30T15:50:51.746537-05:00
Running job bundle output test: /Users/morgane/dev/github/deadline-cloud-for-maya/job_bundle_output_tests/cube

cube
Test succeeded

Timestamp: 2024-10-30T15:50:52.042350-05:00
Running job bundle output test: /Users/morgane/dev/github/deadline-cloud-for-maya/job_bundle_output_tests/layers_no_variation

layers_no_variation
Test succeeded

Timestamp: 2024-10-30T15:50:54.232503-05:00
```

Yes

No
----

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ['3.9', '3.10', '3.11']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
     uses: aws-deadline/.github/.github/workflows/reusable_python_build.yml@mainline
     with:
       os: ${{ matrix.os }}
--- a/depsBundle.py
+++ b/depsBundle.py
@@ -9,7 +9,7 @@ from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any
 
-SUPPORTED_PYTHON_VERSIONS = ["3.7", "3.8", "3.9", "3.10", "3.11"]
+SUPPORTED_PYTHON_VERSIONS = ["3.9", "3.10", "3.11", "3.12"]
 SUPPORTED_PLATFORMS = ["win_amd64", "manylinux2014_x86_64", "macosx_10_9_x86_64"]
 NATIVE_DEPENDENCIES = ["xxhash"]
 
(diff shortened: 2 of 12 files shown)


job_bundle_output_tests/layers/expected_job_bundle/parameter_values.yaml
```diff
@@ -48,7 +48,7 @@ parameterValues:
 - name: RenderSetupIncludeLights
   value: 'true'
 - name: ArnoldErrorOnLicenseFailure
-  value: 'false'
+  value: 'true'
```
AWS-Samuel: This seems to imply we've changed the default error behavior, which I think would be considered a breaking change.

worker.log
AWS-Samuel, line 1: Was including this file intentional?

AWS-Samuel (changes requested):

job_bundle_output_tests/layers/expected_job_bundle/parameter_values.yaml
epmog: Naw, this is just a consequence of our tests. I talked about this in the description

lucaseck (dismissed):

AWS-Samuel (dismissed):

AWS-Samuel (approved):

lucaseck (approved):

PR #200, opened 2024-11-29
feat!: Allow specifying MAYAPY exe via environment

Ahuge:
This matches the logic we have implemented in the `deadline-cloud-for-cinema-4d` package.

The assumption that the executable `mayapy` exists on the PATH doesn't always make sense, often it is easier to specify the version of MAYA to use by explicit path and using an environment variable is an easy way to do this for customers.

We have defined a new environment variable `MAYAPY_EXECUTABLE` which follows the naming from 
[`deadline-cloud-for-cinema-4d`](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py#L298) of `<EXE_NAME>_EXECUTABLE`

Additional functionality to override the mayapy executable that is used via the environment variable.

Yes this was documented in the `README.md` similar to how it was done in the `deadline-cloud-for-cinema-4d` package
No

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,13 @@ This library requires:
 
 ## Versioning
 
-This package's version follows [Semantic Versioning 2.0](https://semver.org/), but is still considered to be in its 
+This package's version follows [Semantic Versioning 2.0](https://semver.org/), but is still considered to be in its
 initial development, thus backwards incompatible versions are denoted by minor version bumps. To help illustrate how
 versions will increment during this initial development stage, they are described below:
 
-1. The MAJOR version is currently 0, indicating initial development. 
-2. The MINOR version is currently incremented when backwards incompatible changes are introduced to the public API. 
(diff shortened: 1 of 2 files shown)


epmog (dismissed):

Ahuge:
Updated to use `MAYA_ADAPTOR_MAYAPY_EXECUTABLE` naming instead of `MAYA_ADAPTOR_MAYAPY_EXE`, which matches Nuke

jericht (dismissed):

Ahuge:
Updated to use MAYAPY_EXECUTABLE naming instead of MAYA_ADAPTOR_MAYAPY_EXE, after a conversation with Morgan and Mark

epmog (approved):

jericht (approved):

PR #207, opened 2025-01-08
fix(submitter): read includeAllLights setting correctly

### LINKED ISSUE #206: Bug: [Submitter] RenderSetupIncludeLights job parameter is inaccurate if the corresponding option variable is not set
### Describe the bug

The `RenderSetupIncludeLights` job parameter in the job bundle generated by the submitter is inaccurate if the corresponding option variable is not set. The current implementation uses the [`optionVar` MEL command](https://help.autodesk.com/cloudhelp/CHS/MayaCRE-Tech-Docs/Commands/optionVar.html#flagremove) to check if the `renderSetup_includeAllLights` option variable is set, however if this variable is not set (e.g. by explicitly interacting with the Render Setup editor menu > Options > "Include all lights in each render layer by default" checkbox), the submitter will always report the option to be `false`, even if the checkbox is checked in Maya. This causes confusion when local renders include all lights, but renders on Deadline Cloud do not.

### Expected Behaviour

`RenderSetupIncludeLights` in the generated job bundle should match the Render Setup editor menu > Options > "Include all lights in each render layer by default" checkbox state

### Current Behaviour

It does not match the state if the `renderSetup_includeAllLights` option variable is not defined.

### Reproduction Steps

1. Open an empty Maya scene
2. In the Render Setup editor menu > Options > "Include all lights in each render layer by default", make sure the option is checked
3. In the MEL prompt, make sure the `renderSetup_includeAllLights` option variable is removed by running `optionVar -remove renderSetup_includeAllLights`
4. Double check that the Render Setup editor menu > Options > "Include all lights in each render layer by default" is still checked (do not touch it, this will set the option variable again)
5. Double double check, in the MEL prompt, make sure the `renderSetup_includeAllLights` option variable does not exist by running `optionVar -ex renderSetup_includeAllLights`. It should return 0
6. Open the Deadline Cloud submitter and export the job bundle
7. Open the job bundle's `parameter_values.yaml` file and verify that the `RenderSetupIncludeLights` option is `false`, which does not match the checkbox value in Maya

### Environment

At minimum:
1. Operating system: All
2. Version of Maya: 2024
3: Version of this package: [0.14.4](https://github.com/aws-deadline/deadline-cloud-for-maya/releases/tag/0.14.4)
    -  Not in my testing (macOS), but it will also happen in that case

jericht:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-maya/issues/206

The IncludeAllLights by default setting for render layers is not being captured correctly by the submitter. The current implementation queries if the corresponding optionVar is set, however if it is not set, then it treats the value as false, even if the actual check box is checked. This value is fed into the `RenderSetupIncludeLights` job parameter which is used by the adaptor to enable/disable the feature for all render layers. This causes confusion when renders submitted to Deadline Cloud return images with incorrect lighting that cannot be reproduced locally since the option is checked locally but not in Deadline Cloud.

Revert the logic for populating the `RenderSetupIncludeLights` job parameter to the [original implementation](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/e39114c6554cbc08977e948192109c8b47b4e57c/src/deadline_submitter_for_maya/scripts/aws_deadline/render_layers.py#L61-L69)

The `RenderSetupIncludeLights` job parameter reflects the current state of the corresponding option

- Ran unit tests
- Manually tested in Maya. Verified that a scene file that has the option checked but has the corresponding option var unset generates a job bundle with RenderSetupIncludeLights set to true.

```

Timestamp: 2025-01-07T17:37:26.535028-08:00
Running job bundle output test: /Users/jericht/Repositories/deadline-cloud-for-maya/job_bundle_output_tests/layers

layers
Test succeeded

Timestamp: 2025-01-07T17:37:27.595248-08:00
Running job bundle output test: /Users/jericht/Repositories/deadline-cloud-for-maya/job_bundle_output_tests/referenced_layers

referenced_layers
Test succeeded

Timestamp: 2025-01-07T17:37:28.054335-08:00
Running job bundle output test: /Users/jericht/Repositories/deadline-cloud-for-maya/job_bundle_output_tests/renderman
Timestamp: 2025-01-07T17:37:28.054420-08:00
Running job bundle output test: /Users/jericht/Repositories/deadline-cloud-for-maya/job_bundle_output_tests/cube

cube
Test succeeded

Timestamp: 2025-01-07T17:37:28.588416-08:00
Running job bundle output test: /Users/jericht/Repositories/deadline-cloud-for-maya/job_bundle_output_tests/layers_no_variation

layers_no_variation
Test succeeded

Timestamp: 2025-01-07T17:37:31.931839-08:00
```

N/A

No

----

### DIFF
--- a/src/deadline/maya_submitter/render_layers.py
+++ b/src/deadline/maya_submitter/render_layers.py
@@ -7,6 +7,7 @@ from contextlib import contextmanager
 from enum import IntEnum
 from typing import List
 
+import maya.app.renderSetup.model.renderSetupPreferences as renderSetupPrefs  # type: ignore
 import maya.cmds
 import maya.mel
 
@@ -64,8 +65,10 @@ def render_setup_include_all_lights() -> bool:
     Returns whether a Render Layer should contain all lights in the scene automatically
     (machine level setting)
     """
-    # The maya.cmds.optionVar query did not work
-    return bool(maya.mel.eval("optionVar -q renderSetup_includeAllLights"))
+    try:
+        return renderSetupPrefs.IncludeAllLightsSetting.isEnabled()
+    except AttributeError:
+        return True
 
 
 @contextmanager


mwiebe (approved):

jairaws (approved):

leongdl (approved):
thank you.

godobyte (approved):

benl-2023 (approved):

epmog:
Because this is a per-machine setting, we should be updating our job bundle output test infrastructure to ensure we can have a test so we don't regress on this. Will require us to change how we set that value on the test machines as we basically want to control this at the test level.

https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/job_bundle_output_test_runner.py#L34-L49

jericht:
> Because this is a per-machine setting, we should be updating our job bundle output test infrastructure to ensure we can have a test so we don't regress on this. Will require us to change how we set that value on the test machines as we basically want to control this at the test level.
> 
> https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/job_bundle_output_test_runner.py#L34-L49

Sounds good, I'll do this in a follow-up PR! https://github.com/aws-deadline/deadline-cloud-for-maya/issues/208

PR #223, opened 2025-01-29
chore: move unit tests under "test/unit" folder

AlexTranAmz:
Want to refactor code to separate unit and integ test
Separate it into different folder
Cleaner code, prepare for adding integ test to the repo

`hatch build && hatch run test`

Did not need to run since this is a refactor

N/A

No

----


jericht (approved):

jericht (commented):
Just double checking, do the [integ/e2e test commands](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/e3ca37dbd614bbf576088deaf35f13d2ee6865d8/hatch.toml#L37-L41) still work?

```
hatch run integ:test
hatch run e2e:test
```

jairaws (approved):

PR #224, opened 2025-01-31
fix: Switch to using `byFrameStep` instead of `byFrame`

### LINKED ISSUE #220: Bug: Override Frame Range
### Describe the bug

Override Frame Range - error when rendering 60fps, must activate the override and change the number after the colon to 1. Defaults to :2 the first time we render in submitter.

### Expected Behaviour

It should respect the Override frame Range as specified after the colon. 

### Current Behaviour

Defaults to :2 the first time we render in submitter.

### Reproduction Steps

- Load a scene in Maya. Open the Deadline Cloud Maya submitter >  Job-Specification settings
- Check Override Frame Range and set the number after the colon. 
- Submit the Job

### Environment

At minimum:
1. Operating system: (e.g. Windows Server 2022; Amazon Linux 2023; etc.) Windows 
2. Version of Maya: 2024
3: Version of this package: DeadlineCloudSubmitter-2025.01.16.0-windows-x64-installer.exe
DeadlineCloudSubmitter-2025.01.16.0-windows-x64-installer.exe

Ahuge:
Fixes: #220 

This is talked about in the following link:
https://help.autodesk.com/view/MAYAUL/2023/ENU/?guid=__Nodes_renderGlobals_html

Essentially `byFrame` is no longer used but kept for backwards compatibility. And the value it gives us is different than `byFrameStep`

Switch to `byFrameStep`
Frame rates working as expected
Using the scene files provided in #220

No

No

----

### DIFF
--- a/src/deadline/maya_submitter/scene.py
+++ b/src/deadline/maya_submitter/scene.py
@@ -56,7 +56,7 @@ class Animation:
         """
         Returns the frame step of the current render.
         """
-        return int(maya.cmds.getAttr("defaultRenderGlobals.byFrame"))
+        return int(maya.cmds.getAttr("defaultRenderGlobals.byFrameStep"))
 
     @staticmethod
     def extension_padding() -> int:


jairaws:
Thanks for the fix. There's an issue with the commit description. The semantic check is failing because of it. I'll check out the change and publish another commit to fix it.

benl-2023 (approved):

jairaws (approved):

PR #227, opened 2025-02-03
chore: add installer descption to support Maya 2025

AlexTranAmz:
Need to indicate we support Maya 2025 in the installer

Add text to show we support Maya 2025
Customer know we have Maya 2025 support
`hatch run test`

No

----


leongdl (approved):

jericht (approved):

jairaws (commented):
This should be a `doc` update, not `chore`.

PR #229, opened 2025-02-04
fix: Load scene settings when the submitter is open

### LINKED ISSUE #167: Bug: Submitter not updating render settings after saving scene
### Expected Behaviour

A Maya scene is opened and the integrated submitter for Deadline Cloud is opened. User closes the integrated submitter window, goes back to the scene and make changes to the render settings (add a new camera and change image size), saves the scene, then opens the integrated submitter for Deadline Cloud again. The changes to the render settings should be reflected in the integrated submitter after saving the scene.

### Current Behaviour

A Maya scene is opened and the integrated submitter for Deadline Cloud is opened. User closes the integrated submitter window, goes back to the scene and make changes to the render settings (add a new camera and change image size), saves the scene, then opens the integrated submitter for Deadline Cloud again. The new renderable camera is not listed as an option in the integrated submitter. After submitting the scene to Deadline Cloud after opening the submitter the second time, it shows that the image size was not updated in the job parameters either. These changes will only be reflected in the submitter if Maya is closed and reopened, or if the scene is saved under a different name. 

First image shows the newly added camera1 is listed in the render settings, but not listed in the submitter after saving the scene. Second image shows the job parameters of the job submitted after saving the scene in the Deadline Cloud monitor beside the Maya render settings.
Screenshot 2024-08-08 at 3 33 30 PM
image

### Reproduction Steps

Recreated using Maya 2024 on Windows and MacOS, using [current version](https://github.com/aws-deadline/deadline-cloud-for-maya/tree/aece21fc5ff7ee9289b22052389a1407ae4ff736) of `deadline-cloud-for-maya`. 
1. Open a Maya scene
2. Open the Deadline Cloud integrated submitter from within Maya (does not matter whether you submit a job or not)
3. Close the integrated submitter window
4. Make changes to the scene. Add a camera, open the render settings window and add that new camera as a renderable camera, and change the image size. 
5. Save the scene.
6. Open the Deadline Cloud integrated submitter again to see if the camera list has been updated.
7. Submit the job from the submitter and look at the job parameters in the Deadline Cloud monitor to see if the image dimensions have been updated.

### Code Snippet

Tried running the commands of `get_renderable_camera_names()` from [cameras.py](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/cameras.py#L13) in Maya's script editor which returned the correct list of renderable cameras. 

`get_renderable_camera_names()` is called from within `show_maya_render_submitter()` in [maya_render_submitter.py](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/maya_render_submitter.py#L445)

`show_maya_render_submitter()` is created in [line 66](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/mel_commands.py#L66) of mel_commands.py. From adding `logger.info()` to each section of the `if else` block starting on [line 63](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/mel_commands.py#L63), it appears that when the integrated submitter is opened for the first time, it enters the `else` block, but after closing, saving the scene and reopening, it enters the `if` block despite the submitter being closed. This means, it does not make a call to `show_maya_render_submitter()` which would fetch the new render settings.

jairaws:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-maya/issues/167

The submitter setting are only updated when Maya is opened or the scene name changes.

Update the submitter settings whenever the submitter is opened.

Improved UX.

* Replicated the reported issue.
* Checked that the submitter settings are updated when opened.
* Submitted job and verified it completed successfully.

Yes

No, the test button kept saying that the module wasn't installed. However, the submitter worked with the regular button.

N/A

No

----

### DIFF
--- a/src/deadline/maya_submitter/mel_commands.py
+++ b/src/deadline/maya_submitter/mel_commands.py
@@ -59,12 +59,18 @@ class DeadlineCloudSubmitterCmd(om.MPxCommand):
                         DeadlineCloudSubmitterCmd.dialog.close()
                     DeadlineCloudSubmitterCmd.dialog = None
 
-                # Show existing submitter dialog, or create new one if needed
+                # Create a new submitter dialog. If this is the first time the submitter is
+                # opened, load the sticky settings. If this is not the first time, close
+                # the existing dialog and create a new one without loading the sticky
+                # settings.
                 if DeadlineCloudSubmitterCmd.dialog:
-                    DeadlineCloudSubmitterCmd.dialog.show()
+                    DeadlineCloudSubmitterCmd.dialog.close()
+                    DeadlineCloudSubmitterCmd.dialog = show_maya_render_submitter(
+                        parent=mainwin, f=Qt.Tool, load_sticky_setting=False
+                    )
                 else:
                     DeadlineCloudSubmitterCmd.dialog = show_maya_render_submitter(
-                        parent=mainwin, f=Qt.Tool
+                        parent=mainwin, f=Qt.Tool, load_sticky_setting=True
                     )
                     DeadlineCloudSubmitterCmd.dialog_scene_name = scene_name
 
--- a/test/unit/deadline_submitter_for_maya/test_mel_commands.py
+++ b/test/unit/deadline_submitter_for_maya/test_mel_commands.py
@@ -1,0 +1,37 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+from deadline.maya_submitter.mel_commands import DeadlineCloudSubmitterCmd
+from unittest.mock import patch, Mock
+import maya.api.OpenMaya as om  # pylint: disable=import-error
+from qtpy.QtCore import Qt  # type: ignore
+from qtpy.QtWidgets import (  # type: ignore
+    QApplication,
+)
+
+
+@patch("deadline.maya_submitter.mel_commands.show_maya_render_submitter")
+@patch.object(QApplication, "instance")
+@patch.object(om.MGlobal, "mayaState")
+def test_deadline_cloud_submitter_cmd_sticky_setting_load_only_once(
+    mock_maya_state: Mock, mock_q_app: Mock, mock_show_maya_render_submitter: Mock
+) -> None:
+    mock_maya_state.return_value = om.MGlobal.kInteractive
+    maya_widget: Mock = Mock()
+    maya_widget.objectName.return_value = "MayaWindow"
+    q_app_instance: Mock = Mock()
+    q_app_instance.topLevelWidgets.return_value = [maya_widget]
+    mock_q_app.return_value = q_app_instance
+    submitter_dialog: Mock = Mock()
+    mock_show_maya_render_submitter.return_value = submitter_dialog
+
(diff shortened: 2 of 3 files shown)


leongdl (dismissed):

src/deadline/maya_submitter/mel_commands.py
```diff
@@ -61,10 +61,12 @@ def doIt(_):  # pylint: disable=invalid-name,
 
                 # Show existing submitter dialog, or create new one if needed
                 if DeadlineCloudSubmitterCmd.dialog:
-                    DeadlineCloudSubmitterCmd.dialog.show()
+                    DeadlineCloudSubmitterCmd.dialog = show_maya_render_submitter(
+                        parent=mainwin, f=Qt.Tool, load_sticky_setting=False
+                    )
```
jericht: It looks like we create a new dialog each time here instead of reusing the existing one. Any issues with that?

sakshie95: +1, it looks like we changed the expected ``# Show existing submitter dialog, or create new one if needed`` in the if block here?

test/unit/deadline_submitter_for_maya/test_mel_commands.py
```diff
@@ -0,0 +1,34 @@
...
+from qtpy.QtCore import Qt  # type: ignore
+from qtpy.QtWidgets import (  # type: ignore
+    QApplication,
+)
+
+
+@patch("deadline.maya_submitter.mel_commands.show_maya_render_submitter")
+@patch.object(QApplication, "instance")
+@patch.object(om.MGlobal, "mayaState")
+def test_deadline_cloud_submitter_cmd_sticky_setting_load_only_once(
```
sakshie95, line 14: Do we need another test for validating the scene name change that should change not load the sticky settings that's called out in the PR description? So we don't break it later

src/deadline/maya_submitter/mel_commands.py
jericht: It looks like I am now able to create multiple submitter dialog windows by clicking the button multiple times vs. before it would only show one window. Maybe we should `.close` the previous dialog before opening a new one? I think having multiple of the same window would be a confusing UX when only one of them is needed

jairaws: Great catch! Will take a look.

test/unit/deadline_submitter_for_maya/test_mel_commands.py
jairaws: The piece of code regarding the scene name change is beyond the scope of this PR.

leongdl (approved):

test/unit/deadline_submitter_for_maya/test_mel_commands.py
sakshie95: Oh I see its a part of the problem statement but not the solution. I assumed ``render_settings.load_sticky_settings(Scene.name())`` does the work for us to not load sticky settings for a new scene name

sakshie95 (approved):

PR #231, opened 2025-02-05
fix: OutputPath is read from the scene in the submitter

jairaws:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-maya/pull/231

The submitter doesn't update the output file when the project path changes.

Modify the Scene class to get the root folder when retrieving the project file.

Submitter updates the output path setting correctly.

* Added unit tests.
* Submitted job and verified it succeeded.

Yes. Three tests passed. The renderman test failed, but seems to be an issue with the environment. The test execution fails complaining about dependencies.
```
Timestamp: 2025-02-10T06:07:32.670548+00:00
Running job bundle output test: C:\Users\rdp\workplace\deadline-cloud-for-maya\job_bundle_output_tests\cube

cube
Test succeeded

Timestamp: 2025-02-10T06:07:33.116203+00:00
Running job bundle output test: C:\Users\rdp\workplace\deadline-cloud-for-maya\job_bundle_output_tests\layers

layers
Test succeeded

Timestamp: 2025-02-10T06:07:33.535561+00:00
Running job bundle output test: C:\Users\rdp\workplace\deadline-cloud-for-maya\job_bundle_output_tests\layers_no_variation

layers_no_variation
Test succeeded

Timestamp: 2025-02-10T06:07:33.803405+00:00
Running job bundle output test: C:\Users\rdp\workplace\deadline-cloud-for-maya\job_bundle_output_tests\referenced_layers
```

N/A.

No.

Screenshot 2025-02-05 at 11 23 43 AM
----

### DIFF
--- a/src/deadline/maya_submitter/scene.py
+++ b/src/deadline/maya_submitter/scene.py
@@ -130,7 +130,7 @@ class Scene:
         """
         Returns the base path to the project the current scene is in
         """
-        return maya.cmds.workspace(query=True, directory=True)
+        return maya.cmds.workspace(query=True, rootDirectory=True)
 
     @staticmethod
     def output_path() -> str:
@@ -140,9 +140,9 @@ class Scene:
         # This one didn't work translated to the maya.cmds equivalent
         image_rule = maya.mel.eval('workspace -q -fileRuleEntry "images"')
         if image_rule:
-            return os.path.join(maya.cmds.workspace(query=True, directory=True), image_rule)
+            return os.path.join(Scene.project_path(), image_rule)
         else:
-            return maya.cmds.workspace(query=True, directory=True)
+            return Scene.project_path()
 
     @staticmethod
     def autotx() -> bool:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.15.1 (2025-02-03)
+
+* This release only includes documentation updates (#227) ([`e0a3827`](https://github.com/aws-deadline/deadline-cloud-for-maya/commit/e0a38272e6cab7dac27f0af8c7c64ce3a79189ae))
+
+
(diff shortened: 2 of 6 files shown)


AlexTranAmz (approved):

jericht:
> Did you run the "Job Bundle Output Tests"? If not, why not? If so, paste the test results here.
> No, can't not open the job bundle out tests dialog.

Do you have `DEADLINE_ENABLE_DEVELOPER_OPTIONS` set to `True` in your shell? I think we should run these tests

src/deadline/maya_submitter/scene.py
```diff
@@ -140,9 +140,9 @@ def output_path() -> str:
         # This one didn't work translated to the maya.cmds equivalent
         image_rule = maya.mel.eval('workspace -q -fileRuleEntry "images"')
         if image_rule:
-            return os.path.join(maya.cmds.workspace(query=True, directory=True), image_rule)
+            return os.path.join(Scene.project_path(), image_rule)
```
leongdl, line 143: OK so we are just getting `rootDirectory` instead of `directory`

leongdl (approved):

jairaws:
> > Did you run the "Job Bundle Output Tests"? If not, why not? If so, paste the test results here.
> > No, can't not open the job bundle out tests dialog.
> 
> Do you have `DEADLINE_ENABLE_DEVELOPER_OPTIONS` set to `True` in your shell? I think we should run these tests

I was able to run the test. I had an issue with the env variable first, then an issue with Renderman. Three of the tests passed, but the `referenced_layers` throws an exception due to the Renderman configuration. We need to fix the test logic to skip this test if Renderman isn't installed, or make it a requirement and document it.

Screenshot 2025-02-09 at 10 05 46 PM

PR #236, opened 2025-02-07
feat: Arnold licensing error for mtoa5.4.7.1

ZainAallii:
Fixes: *NA*

A customer reported an issue where their Maya job succeeds even when abort_on_license_fail option is enabled and license checkout for Arnold fails. This is happening because the string which we are trying to match in adaptor.py line#217 is different from what is actually printing in the log. What was the solution? (How)
I changed the string to catch both the error messages:
- aborting render because the abort_on_license_fail option was enabled
- borting render because this is a batch render and abort_on_license_fail option is enabled

It will catch Arnold license error an fail the job.

I ran the unit test by running `hatch run all:test`, I got `Required test coverage of 41.0% reached. Total coverage: 44.05%`

Yes. The result is here: 
[MayaUnitTest.txt](https://github.com/user-attachments/files/18714996/MayaUnitTest.txt)

No, I ran the adaptor locally and tested it with real scene file. I did not have any licenses for Arnold so my job failed on the error message. 
MayaAdaptor
Check the logs: 
[MayaAdaptorLocalRun.txt](https://github.com/user-attachments/files/18714999/MayaAdaptorLocalRun.txt)

NA

No
----


test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py
```diff
@@ -866,7 +866,9 @@ def test_error_on_arnold_license_fail(
         init_data["error_on_arnold_license_fail"] = error_on_arnold_license_fail
         adaptor = MayaAdaptor(init_data)
         expected_regex_list = [
-            re.compile("(aborting render because the abort_on_license_fail option was enabled)")
+            re.compile(
```
jairaws, line 870: This unit test only checks that the expected regexes are in the list of callback regexes. Do we have a unit test to check that the error message is caught for all cases? If so, we should update that unit test to include the new case. If not, we should add a unit test to validate that the error messages are caught.

ZainAallii: I did test it on with Maya Arnold job (MTOA5.4.7.1) manually. Please check the logs attached https://github.com/user-attachments/files/18714999/MayaAdaptorLocalRun.txt

jairaws: Synced up with the requester offline. There's some reluctance to add the unit test based on time concerns. However, I believe we should promote high quality standards for our code. Thus, IMO, we should add test coverage for every behaviour change, even if there wasn't a test for similar behaviour before.

I wrote the test myself to help unblock this PR:
```
from typing import List
...
    @pytest.mark.parametrize(
        "message, should_match",
        [
            ("aborting render because the abort_on_license_fail option was enabled", True),
            (
                "aborting render because this is a batch render and abort_on_license_fail option is enabled",
                True,
            ),
            ("aborting render because ", False),
            ("aborting render because something else happened", False),
        ],
    )
    def test_regex_for_error_on_arnold_license_fail(
        self, message: str, should_match: bool
    ) -> None:
        """
        Tests that the regexes used to catch arnold license errors work as expected.
        """
        # GIVEN
        regex_list = [
            re.compile(
                "(aborting render because (?:the abort_on_license_fail option was enabled|this is a batch render and abort_on_license_fail option is enabled))"
            )
        ]

        # WHEN
        results: List[bool] = [regex.match(message) for regex in regex_list]

        # THEN
        assert any(results) == should_match
```

@ZainAallii please add this test and update the PR. Feel free to modify it or add any other test case.

ZainAallii: Hello @jairaws 
I added the test as suggested. 
[test.txt](https://github.com/user-attachments/files/19171892/test.txt)

jairaws: Thank you!

.github/ISSUE_TEMPLATE/maintenance.yml
```diff
@@ -1,7 +1,7 @@
 name: "🛠️ Maintenance"
 description: Some type of improvement
 title: "Maintenance: (short description of the issue)"
-labels: ["needs triage"]
+labels: ["maintenance", "needs triage"]
```
jairaws, line 4: Something seems off with you branch. Some of these changes are from other PRs. Did you `pull --rebase` your commit before pushing it?

PR #255, opened 2025-03-25
feat: Add standalone installer for Maya Submitter

AlexTranAmz:
We want to separate submitter installer so each DCC will have a standalone installer. This will help with releasing independently
Add code/workflow/test to create a standalone installer
Customer/developer can now create an standalone installer for just Maya
`hatch run test-installer`

Not applicable in this case

yes

No

----


jairaws (dismissed):

jairaws (commented):
I tested the installers from this PR in Linux and Windows.

DEVELOPMENT.md
```diff
@@ -48,6 +52,7 @@ from any directory of this repository:
 * `hatch env prune` - Delete all of your isolated workspace [environments](https://hatch.pypa.io/1.12/environment/)
    for this package.
 * `hatch run install` - Install the DeadlineCloudForMaya plugin from this repository into a temporary directory within this repository.
+* `hatch run build-installer` - To build the submitter installer for Maya into the root directory
```
miabatta: Nit: believe this will have to be `hatch run installer:build-installer`. Same for lines 269 and 272

test/installer/test_installer.py
```diff
@@ -0,0 +1,530 @@
...
+    assert uninstaller in top_level_dir
+
+    # Just check that we have dependencies in this folder
+    module_dir = [f.name for f in (python_dir).iterdir()]
+    assert "deadline" in module_dir
+    assert "qtpy" in module_dir
+    assert "xxhash" in module_dir
+
+    # Check the Maya module is here and there's a version file
+    # Check the Maya module is here and there's a version file
```
miabatta: nit: duplicate comment

miabatta (dismissed):
Thanks a bunch! Two documentation nits that you can fix here or in a later PR.

DEVELOPMENT.md
AlexTranAmz: Fixed

test/installer/test_installer.py
AlexTranAmz: Fixed

miabatta (approved):

jairaws (approved):

PR #259, opened 2025-04-05
Better job attachments handling to include cache files

### LINKED ISSUE #256: Bug: Submitter and Adaptor don't account for Cache Files
### Describe the bug

If you use the MASH plugin or the Bifrost plugin and have an effect with some type of simulation or randomization, the submitter will not add the cache for those effects to the job attachments and the adaptor will be unable to pathmap those files if added manually.

Related: https://github.com/aws-deadline/deadline-cloud-for-maya/issues/245

### Expected Behaviour

I expected cache files to be treated like any other dependency

### Current Behaviour

Cache files are ignored entirely. If you're using a simulation, each render node will have to perform the simulation for every frame which drastically slows down render times. If you're performing a randomization, the render nodes might produce inconsistent results between rendered frames.

### Reproduction Steps

Create a new Maya project.

Add a simulation or randomization.

Save it to cache and try to render.

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.)
2. Version of Maya
3. Version of this package

yuanmich2:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-maya/issues/256

Caching files is required for any scene with a simulation.

My solution checks every node and adds any attributes that are file paths to the job attachments list

Any file references get added to the job attachments list

Repeatedly running render tests on my CMF and SMF fleets. Also I ran `hatch run test` and didn't get any failures

It said it successfully ran 4 tests.

No

Not that I know of.

----

### DIFF
--- a/src/deadline/maya_adaptor/MayaAdaptor/adaptor.py
+++ b/src/deadline/maya_adaptor/MayaAdaptor/adaptor.py
@@ -47,6 +47,7 @@ _MAYA_INIT_KEYS = {
     "output_file_prefix",
     "render_layer",
     "render_setup_include_lights",
+    "cache_pathmapping",
     "error_on_arnold_license_fail",
 }
 
--- a/src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
+++ b/src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
@@ -47,6 +47,7 @@ class DefaultMayaHandler:
             "project_path": self.set_project_path,
             "render_layer": self.set_render_layer,
             "render_setup_include_lights": self.set_render_setup_include_lights,
+            "cache_pathmapping": self.set_cache_pathmapping,
             "scene_file": self.set_scene_file,
         }
         self.image_width = None
@@ -206,6 +207,21 @@ class DefaultMayaHandler:
         DirectoryMapping.set_activated(True)
         for source, dest in rules.items():
             DirectoryMapping.mappings[source] = dest
+
+    def set_cache_pathmapping(self, data: dict) -> None:
+        """
+        Applies pathmapping to some cache files. Some nodes convert their cache file paths into absolute paths and store them in an attribute named 'absoluteCachePath'. This attribute is only updated when the UI does a refresh, which means when we render headlessly, the attribute doesn't get changed properly. This function searches for nodes with the 'absoluteCacheName' attribute and manually changes it to the correct value.
+
+        Args:
+            data (dict): The data given from the Adaptor. Keys expected: []
+        """
+        if not DirectoryMapping.get_activated():
+            return
+        for node in maya.cmds.ls():
+            if maya.cmds.attributeQuery("absoluteCacheName", node=node, exists=True):
+                attrName: str = "%s.absoluteCacheName" % node
+                new_cache_path: str = DirectoryMapping.convert(maya.cmds.getAttr(attrName))
+                maya.cmds.setAttr(attrName, new_cache_path)
 
     def set_project_path(self, data: dict) -> None:
         """
--- a/src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
+++ b/src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
@@ -210,7 +210,9 @@ class DefaultMayaHandler:
 
     def set_cache_pathmapping(self, data: dict) -> None:
         """
-        Applies pathmapping to some cache files. Some nodes convert their cache file paths into absolute paths and store them in an attribute named 'absoluteCachePath'. This attribute is only updated when the UI does a refresh, which means when we render headlessly, the attribute doesn't get changed properly. This function searches for nodes with the 'absoluteCacheName' attribute and manually changes it to the correct value.
+        Applies pathmapping to some cache files. Some nodes convert their cache file paths into absolute paths and store them in an attribute named 'absoluteCachePath'.
+        This attribute is only updated when the UI does a refresh, which means when we render headlessly, the attribute doesn't get changed properly.
+        This function searches for nodes with the 'absoluteCacheName' attribute and manually changes it to the correct value.
 
         Args:
             data (dict): The data given from the Adaptor. Keys expected: []
@@ -221,7 +223,7 @@ class DefaultMayaHandler:
             if maya.cmds.attributeQuery("absoluteCacheName", node=node, exists=True):
                 attrName: str = "%s.absoluteCacheName" % node
                 new_cache_path: str = DirectoryMapping.convert(maya.cmds.getAttr(attrName))
-                maya.cmds.setAttr(attrName, new_cache_path)
+                maya.cmds.setAttr(attrName, new_cache_path, type="string")
 
     def set_project_path(self, data: dict) -> None:
         """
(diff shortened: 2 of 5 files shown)


jairaws:
Thanks for publishing this PR!

Some of the automatic checks are failing. The commits need to be signed, and there seems to be a problem with an import reference.

src/deadline/maya_submitter/assets.py
```diff
@@ -46,6 +48,33 @@ def parse_scene_assets(self) -> set[Path]:
...
+                    if frame_re_matches or "<f>" in normalized_path or "<frame>" in normalized_path:
+                        for path in self._expand_path(normalized_path):
+                            if not os.path.exists(path):
+                                continue
+                            if path in assets:
+                                continue
+                            assets.add(path)
+                    elif os.path.exists(normalized_path) and normalized_path not in assets:
+                        assets.add(Path(normalized_path))
+                except Exception as e:
```
jairaws: what part of this code do we expect to throw an exception?

src/deadline/maya_submitter/assets.py
```diff
@@ -46,6 +48,33 @@ def parse_scene_assets(self) -> set[Path]:
             for path in self._expand_path(normalized_path):
                 assets.add(path)
 
+        # Iterate through every node and list all attributes that are filenames
```
jairaws, line 72: Could we create a function to retrieve all path files? That would make the code cleaner and more maintainable. Also, could we breakdown this functionality into independent functions. This nested structure adds complexity to the code. We can model it like:

nodes -> filter_nodes_without_attributes() -> filter_empty_attributes() -> normalize_path_files() -> all_path_files

src/deadline/maya_submitter/assets.py
```diff
@@ -156,7 +185,7 @@ def _expand_path(self, path: str) -> Generator[Path, None, None]:
         frame_re_matches = _FRAME_RE.findall(path)
 
         frame_list: Iterable[int] = [0]
-        if frame_re_matches or "<f>" in path:
+        if frame_re_matches or "<f>" in path or "<frame>" in path:
```
jairaws, line 323: Why is `<frame>` needed here? Can we add a comment?

src/deadline/maya_submitter/assets.py
yuanmich2: Maya throws an exception if you try to get an attribute with a dot in it. I haven't figured out what the dot means or how to properly handle it, so I'm just ignoring it for now. I think it means the attribute is a nested attribute though. It'll be something like "node.transform.position" instead of just "node.position"

src/deadline/maya_submitter/assets.py
yuanmich2: I agree that it could be a little bit more readable if I broke it up into smaller functions. But I propose this flow:
get_all_file_attributes() -> expand() -> normalize() -> all_paths

src/deadline/maya_submitter/assets.py
yuanmich2: Update: I figured out how to handle it. You need to use the "multi=true" option when querying attributes to get the correct one.

src/deadline/maya_submitter/assets.py
yuanmich2: frame is needed because it's the same as "\<f\>" but we weren't handling it. Also the default filename that Maya uses for cache file names includes "\<frame\>" not "\<f\>"

src/deadline/maya_submitter/assets.py
yuanmich2: The new flow has been implemented in the latest version

yuanmich2:
The parsing of every attribute in the scene might have performance impacts on opening the submitter and the worker loading the scene for the first time, so we should investigate that and/or find a way to make this a toggleable feature

src/deadline/maya_adaptor/MayaAdaptor/adaptor.py
```diff
@@ -47,6 +47,7 @@ class MayaNotRunningError(Exception):
     "output_file_prefix",
     "render_layer",
     "render_setup_include_lights",
+    "cache_pathmapping",
```
leongdl, line 50: https://github.com/aws-deadline/deadline-cloud-for-maya/pull/296

Rebased for Michael, please dis-regard this PR.

jericht:
In my testing, caching cut down render time by about half (the performance boost probably depends on the size and complexity of the render, this is just a data point)

Average render time with caching: 6-7 seconds
Average render time without caching: 12-13 seconds

I had to push up a commit to get the adaptor side working, but I'm confident that these changes are working as expected. I don't see any performance regressions on the adaptor side.

The part where we scan the entire node tree happens only on the initialization of the adaptor, so just once at the start of the session, not for each frame.

Before we merge, I want to add some unit tests

jericht:
I've opened up a separate PR (https://github.com/aws-deadline/deadline-cloud-for-maya/pull/308) to backfill unit tests for the existing code that was modified in this PR. Instead of waiting for above to get merged, let's merge this PR and I'll create another PR to backfill the unit tests for this PR.

...or we can proceed without unit tests (like we have been).

src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
```diff
@@ -207,6 +208,23 @@ def set_path_mapping(self, data: dict) -> None:
...
+        This function searches for nodes with the 'absoluteCacheName' attribute and manually changes it to the correct value.
+
+        Args:
+            data (dict): The data given from the Adaptor. Keys expected: []
+        """
+        if not DirectoryMapping.get_activated():
+            return
+        for node in maya.cmds.ls():
+            if maya.cmds.attributeQuery("absoluteCacheName", node=node, exists=True):
+                attrName: str = "%s.absoluteCacheName" % node
```
AlexTranAmz, line 224: Nit: could we change this to `attrName: str = f"{node}.absoluteCacheName"` for more understandable code?

src/deadline/maya_submitter/assets.py
```diff
@@ -253,6 +267,35 @@ def _get_arnold_texture_files(self):
...
+        for node in maya.cmds.ls():
+            attrs: list[str] = maya.cmds.listAttr(
+                node, usedAsFilename=True, fullNodeName=True, multi=True
+            )
+            # We need to make sure we're including Bifrost caches, but listAttr won't find them
+            # Bifrost simulation caches use the "absoluteCacheName" attribute, which is not marked with "usedAsFilename"
+            if maya.cmds.attributeQuery("absoluteCacheName", node=node, exists=True):
+                if attrs is None:
+                    attrs = []
+                attrs.append("%s.absoluteCacheName" % node)
```
AlexTranAmz, line 288: Same as above

AlexTranAmz (approved):

leongdl (approved):

src/deadline/maya_submitter/assets.py
```diff
@@ -67,6 +69,18 @@ def parse_scene_assets(self, progress_callback=None) -> set[Path]:
                 if progress_callback:
                     progress_callback(f"Processed {i+1}/{total_refs} file references...")
 
+        # Iterate through every node and list all attributes that are filenames
+        # Then replace all tokens and check if it's a real path
+        # If it's not already in assets, add it to assets
+        for path in self._get_node_attr_paths(expand_tokens=True):
+            normalized_path = os.path.normpath(path)
+            frame_re_matches = _FRAME_RE.findall(normalized_path)
+            if frame_re_matches or "<f>" in normalized_path or "<frame>" in normalized_path:
```
Kikkopanda, line 78: Does not cover the case for if an artist defines the number padding in the frame token ie <frame04>

jericht: Thanks for calling this out! This is what is mentioned in https://github.com/aws-deadline/deadline-cloud-for-maya/issues/351 right?

Kikkopanda: yes

PR #260, opened 2025-04-05
test: add minimal integ test for Maya adaptor and submitter

AlexTranAmz:
We want to have integ test to verify the functionality of the adaptor and the submitter every time we make changes to the package

Refactor code, add the infrastructure and minimal test as an example to follow

#### Refactor code
- We changed the code structure to move call back function which responsible for generating job bundle from inside the submitter dialog itself to outside for our test to call. This also bring some refactor to expose certain variable to be called outside the scope of the function
- This help integ test able to generate a job bundle for comparison but also preserve the functionality of the submitter dialog itself.
- We've also generate a custom submitter with this code change and tested it to make sure there is no regression. 

We will now have integ test which save time and less error prone compare to manual testing

`hatch run test && hatch run integ:test`

Yes

No

----


jairaws:
Can we explain in the PR description all the refactoring done?

test/integ/helpers/image_comparison.py
```diff
@@ -0,0 +1,23 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from pathlib import Path
+
+import PIL.Image
+
+import numpy as np
+
+
+def assert_all_images_close(expected_image_directory: Path, actual_image_directory):
```
jairaws: Can we add a comment to this function?

jairaws: `actual_image_directory` is missing a type.

jairaws: The function is missing a return type.

jairaws: I feel that the name of function is not clear. We are comparing to two images, with some tolerance. So the name be something like `are_images_similar`. 

I would also prefer for the function to return a boolean, instead of asserting within. That would increase its reusability. 

Finally, can the tolerance value be a parameter of the function? We can default it to 2.

test/integ/helpers/image_comparison.py
```diff
@@ -0,0 +1,23 @@
...
+            continue
+
+        # Open the two image files with Pillow https://pillow.readthedocs.io/en/stable/index.html
+        # and put them in numpy arrays. Pillow doesn't have a good built-in way to do image comparison
+        # with tolerance.
+        actual = np.asarray(PIL.Image.open(actual_image_directory / image.name))
+        expected = np.asarray(PIL.Image.open(image))
+
+        # Check that the two images are the same within a tolerance.
+        # It's normal for there to be noise in an output image, so it is unlikely that two
```
jairaws: nit: truncated

test/integ/test_maya_adaptors.py
```diff
@@ -0,0 +1,39 @@
...
+
+import pytest
+from flaky import flaky
+
+from .helpers.image_comparison import assert_all_images_close
+from .helpers.test_runners import run_adaptor_test
+
+
+@pytest.mark.adaptor
+@flaky(max_runs=3, min_passes=1)
```
jairaws, line 21: why is this test flaky?

test/integ/test_maya_submitters.py
```diff
@@ -0,0 +1,212 @@
...
+        scene_location_marker: Mark = request.keywords["scene_files"]
+
+        scene_file: Path
+        for scene_file in scene_location_marker.args:
+            sticky_settings_location = scene_file.with_name(
+                f"{scene_file.stem}.deadline_render_settings.json"
+            )
+            Path(script_location / sticky_settings_location).unlink(missing_ok=True)
+
+        maya.standalone.uninitialize()
```
jairaws: why do we need to `uninitialize` maya at this point?

test/integ/test_maya_submitters.py
```diff
@@ -0,0 +1,212 @@
...
+        with open(job_history_dir / "parameter_values.yaml") as actual:
+            actual_parameter_values = yaml.safe_load(actual)
+            # Compare the lengths before we turn it into a set so that we can cover the case of duplicate parameters.
+            assert len(actual_parameter_values["parameterValues"]) == len(
+                expected_parameter_values["parameterValues"]
+            )
+
+            # The order of the list of parameter values doesn't matter,
+            for parameter_value in expected_parameter_values["parameterValues"]:
+                assert parameter_value in actual_parameter_values["parameterValues"]
```
jairaws: we are just asserting the value exists, but not that the values are equal.

test/integ/test_maya_submitters.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def assert_parameter_values(
+        self, job_history_dir: Path, expected_parameter_values: dict[str, list]
+    ):
+        """
+        Helper function that asserts that parameter values in the job bundle are what's expected.
+        """
+        with open(job_history_dir / "parameter_values.yaml") as actual:
+            actual_parameter_values = yaml.safe_load(actual)
+            # Compare the lengths before we turn it into a set so that we can cover the case of duplicate parameters.
```
jairaws: Where do we turn it into a set?

test/integ/test_maya_submitters.py
```diff
@@ -0,0 +1,212 @@
...
+            assert len(actual_parameter_values["parameterValues"]) == len(
+                expected_parameter_values["parameterValues"]
+            )
+
+            # The order of the list of parameter values doesn't matter,
+            for parameter_value in expected_parameter_values["parameterValues"]:
+                assert parameter_value in actual_parameter_values["parameterValues"]
+
+    def assert_asset_references(
```
jairaws: Could we simply have a method that compares that two dictionaries are equal? And the use for all the yaml files that we need to compare?

test/integ/test_maya_submitters.py
```diff
@@ -0,0 +1,212 @@
...
+            # Compare the lengths before we turn it into a set so that we can cover the case of duplicate assets.
+            assert len(actual_asset_reference["assetReferences"]["inputs"]["filenames"]) == len(
+                expected_asset_references["assetReferences"]["inputs"]["filenames"]
+            )
+            actual_asset_reference["assetReferences"]["inputs"]["filenames"] = set(
+                actual_asset_reference["assetReferences"]["inputs"]["filenames"]
+            )
+            assert actual_asset_reference == expected_asset_references
+
+    @pytest.mark.scene_files(Path("minimal_test") / "scene" / "test.ma")
```
jairaws: Are duplicating the scene path value. We have a reference in the marker metadata, and we set the values we use in the test below as well. Is the only use of the metadata here to make the scene path available to the clean up fixture? If so, can use the fixture as a function that we call at the beginning of the test with the scene path?

test/integ/helpers/image_comparison.py
AlexTranAmz: will make the change to the function accordingly

test/integ/test_maya_adaptors.py
AlexTranAmz: Due to the nature of the image renderer engine, 2 pic will not 100% alike. We need to have a balance of tolerance to make it tight enough but not introduce false positive. 
The tolerance of 2 in the above function make this test sometime fail due to tight tolerance so I mark it as flaky.

test/integ/test_maya_submitters.py
AlexTranAmz: This can be treat as clean up. We want to uninitialize Maya here to:
1. Not accidentally make change to the scene file
2. Make sure we do not keep Maya open if not needed (i.e. adaptor test)

test/integ/test_maya_submitters.py
AlexTranAmz: the parameter_value here is a dict with name and value so we assert both.

test/integ/test_maya_submitters.py
AlexTranAmz: the line right below where we basically iterate through the dict, turning each entry into a set

test/integ/test_maya_submitters.py
AlexTranAmz: I've think about that but due to the nature of these yaml have different dict structure, it will be more complicated to combine it into a common function.
As you can see, most of these is actually going through the dict structure to get the correct set. The common part is comparing set is just a 2-3 lines code

test/integ/test_maya_submitters.py
AlexTranAmz: I think what I will do is make sure the test use the maker metadata to simplify stuff

test/integ/test_maya_submitters.py
```diff
@@ -0,0 +1,205 @@
...
+                    "directories": [str(output_path)],
+                },
+                "referencedPaths": [],
+            }
+        }
+
+        self.assert_asset_references(job_history_dir, expected_asset_references)
+
+        # Clean up after test
+        self._cleanup_sticky_settings(scene_location, script_location)
```
jairaws: Maybe we should actually call this function at the beginning of the test, to clean up sticky setting just in case. With the current implementation, if the test fails, the sticky settings won't be cleaned up.

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -506,6 +505,25 @@ def show_maya_render_submitter(
...
+def on_create_job_bundle_callback(
+    widget: SubmitJobToDeadlineDialog,
+    job_bundle_dir: str,
+    settings: RenderSubmitterUISettings,
+    queue_parameters: list[dict[str, Any]],
+    asset_references: AssetReferences,
+    host_requirements: Optional[dict[str, Any]] = None,
+    purpose: JobBundlePurpose = JobBundlePurpose.SUBMISSION,
+) -> None:
+    with open(Path(__file__).parent / "default_maya_job_template.yaml") as fh:
```
benl-2023, line 520: nit: "default_maya_job_template.yaml" could be a constant

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -506,6 +505,25 @@ def show_maya_render_submitter(
     # Sort the layers by name
     render_layers.sort(key=lambda layer: layer.display_name)
 
+    return render_layers
+
+
+def on_create_job_bundle_callback(
```
benl-2023, line 511: This function is > 100 lines long. Can we make this simpler by extracting some functionality into other classes/modules?

test/integ/test_maya_adaptors.py
benl-2023: Is there an alternative way to compare the image here that wouldn't require a test re-run? Can we do some image comparison that passes when images are 98% alike?

test/integ/test_maya_submitters.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+
+    def _cleanup_sticky_settings(self, scene_file: Path, script_location: Path):
+        """
+        We need to clean the sticky settings before the test runs so that we can ensure
+        a clean environment.
+        """
+
+        sticky_settings_location = scene_file.with_name(
+            f"{scene_file.stem}.deadline_render_settings.json"
```
benl-2023, line 52: nit: could use a constant here.

test/integ/test_maya_submitters.py
```diff
@@ -0,0 +1,205 @@
...
+        )
+        Path(script_location / sticky_settings_location).unlink(missing_ok=True)
+
+    def assert_parameter_values(
+        self, job_history_dir: Path, expected_parameter_values: dict[str, list]
+    ):
+        """
+        Helper function that asserts that parameter values in the job bundle are what's expected.
+        """
+        with open(job_history_dir / "parameter_values.yaml") as actual:
```
benl-2023: nit, could use a constant here

test/integ/test_maya_adaptors.py
AlexTranAmz: Sync offline, the approach of having tolerance as a var is good enough at this point and we decided to keep the flaky there.

pyproject.toml
```diff
@@ -137,6 +137,10 @@ pythonpath = [
   # The Maya submitter plugin isn't structured like a Python package
   "maya_submitter_plugin/plug-ins"
 ]
+markers = [
+    "submitter: mark test as test for the submitter",
```
jairaws, line 141: nit: These descriptions feel too verbose and not too relevant. Is this text used anywhere?

src/deadline/maya_submitter/maya_render_submitter.py
jairaws: +1 It can be done in a separate PR.

jairaws (dismissed):

pyproject.toml
AlexTranAmz: Yes to make sure we do not have warning from pytest for unknown markers

test/integ/helpers/output_comparison.py
```diff
@@ -0,0 +1,78 @@
...
+import yaml
+
+import PIL.Image
+
+import numpy as np
+
+from test.integ.test_const import ASSET_REFERENCES, PARAMETER_VALUES
+
+
+# Function to compare if the render output from the adaptor match what we expected
```
jairaws: Please use the python commenting standard. Ex:
```
def are_images_similar(
    expected_image_directory: Path, actual_image_directory: Path, tolerance: int
) -> bool:
"""
Function to compare if the render output from the adaptor match what we expected.
"""
```

benl-2023 (dismissed):

test/integ/helpers/output_comparison.py
```diff
@@ -0,0 +1,79 @@
...
+        # Check that the two images are the same within a tolerance.
+        # It's normal for there to be noise in an output image, so it is unlikely that two
+        # renders will be exactly the same.
+        if not np.allclose(actual, expected, atol=tolerance):
+            return False
+    return True
+
+
+# Helper function to extract parameter value for path convert if needed
+def extract_parameter_value(expected_parameter_values, param_name):
```
jairaws, line 38: nit: comment not following python standard.

jairaws (approved):

benl-2023 (approved):

PR #271, opened 2025-04-22
fix: fixed Maya submitter to pull the correct image resolution from scene

skarpree:

The Maya submitter doesn't take the changed ImageResolution from the scene, the parameters_values.yaml file from the template shows the wrong ImageWidth and ImageHeight after been changed

Change the code under the maya_render_submitter.py file to pull the Width and Height from Maya scene using the "defaultResolution.height" and "defaultResolution.width". It returned the correct value after this. 

Enables the customers to submit the correct Image Resolution to Deadline for Maya. 

Replaced the maya_render_submitter.py in the local installation and re-run the Maya application. Exporting the scene template from the submitter revealed the correct Image Width and Height in the parameter_values.yaml: 
```
- name: ImageWidth
  value: 320
- name: ImageHeight
  value: 240
```

No.

No.
----

### DIFF
--- a/src/deadline/maya_submitter/maya_render_submitter.py
+++ b/src/deadline/maya_submitter/maya_render_submitter.py
@@ -368,13 +368,13 @@ def _get_parameter_values(
         parameter_values.append(
             {
                 "name": "ImageWidth",
-                "value": render_layers[0].image_resolution[0],
+                "value": get_width(),
             }
         )
         parameter_values.append(
             {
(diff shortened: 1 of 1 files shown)


leongdl (approved):
Thank you for the contribution!

sakshie95 (approved):

benl-2023 (approved):

godobyte (approved):

Issue #239, opened 2025-02-11
Bug: missing adaptor environment enter and exit timeouts


joel-wong-aws:
### Describe the bug

The Maya adaptor's environment enter and exits do not have a timeout specified in their [job template](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/d77f47491372abec3324410e0cd2eb7cbc68a7eb/src/deadline/maya_submitter/default_maya_job_template.yaml#L92-L113).

### Expected Behaviour

The timeout for the environment enter should be a little more (ex. 1 minute) than the [server start timeout + Maya start timeout](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/d77f47491372abec3324410e0cd2eb7cbc68a7eb/src/deadline/maya_adaptor/MayaAdaptor/adaptor.py#L72-L75). The timeout for the environment exit should be a little more (ex. 1 minute) than the [server end timeout + Maya end timeout](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/d77f47491372abec3324410e0cd2eb7cbc68a7eb/src/deadline/maya_adaptor/MayaAdaptor/adaptor.py#L72-L75)

### Current Behaviour

If the environment enter or exit get stuck (and the first layer of timeouts doesn't work), then the worker agent does not enforce the timeout.

### Proposed solution

Add a `timeout` to the environment enter and exit actions [here](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/d77f47491372abec3324410e0cd2eb7cbc68a7eb/src/deadline/maya_submitter/default_maya_job_template.yaml#L92-L113), using the [OpenJD spec `timeout` field](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#5-action).

### Environment

1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.): affects all rendering OSes
2. Version of Maya: affects all versions of Maya
3. Version of this package: up to and including 0.15.1
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? N/A

Example PR: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/173

AlexTranAmz:
Thanks for reporting this issue. We'll take a look to determine the best path forward.

jairaws:
https://github.com/aws-deadline/deadline-cloud-for-maya/pull/274

PR #274, opened 2025-05-02
chore: Set timeouts for the adaptor in the job template

jairaws:
https://github.com/aws-deadline/deadline-cloud-for-maya/issues/239

Set timeouts for the adaptor in the job template 

Add guardrails for better performance.

* `hatch build && hatch run test`
*  Ran test bundle.

No.

No.

----

### DIFF
--- a/src/deadline/maya_submitter/default_maya_job_template.yaml
+++ b/src/deadline/maya_submitter/default_maya_job_template.yaml
@@ -102,6 +102,7 @@ steps:
           - file://{{Env.File.initData}}
           cancelation:
             mode: NOTIFY_THEN_TERMINATE
+          timeout: 87000
         onExit:
           command: MayaAdaptor
           args:
@@ -111,6 +112,7 @@ steps:
           - '{{ Session.WorkingDirectory }}/connection.json'
           cancelation:
             mode: NOTIFY_THEN_TERMINATE
+          timeout: 600
   script:
     embeddedFiles:
     - name: runData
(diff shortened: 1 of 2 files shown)


src/deadline/maya_submitter/default_maya_job_template.yaml
```diff
@@ -102,6 +102,7 @@ steps:
           - file://{{Env.File.initData}}
           cancelation:
             mode: NOTIFY_THEN_TERMINATE
+          timeout: 87000
```
jericht, line 105: How do we choose these values for the timeout?

sakshie95 (approved):

src/deadline/maya_submitter/default_maya_job_template.yaml
jairaws: From the issue requirements:

The timeout for the environment enter should be a little more (ex. 1 minute) than the [server start timeout + 3ds Max start timeout](https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/a14dbf982874ae988c0385eea6054d6a1da47ca1/src/deadline/max_adaptor/MaxAdaptor/adaptor.py#L56-L59). The timeout for the environment exit should be a little more (ex. 1 minute) than the [server end timeout + 3ds Max end timeout](https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/a14dbf982874ae988c0385eea6054d6a1da47ca1/src/deadline/max_adaptor/MaxAdaptor/adaptor.py#L56-L59)

jericht (approved):

PR #281, opened 2025-06-04
fix: sdist failed to install

epmog:

When running `pip install` with an sdist (compressed or uncompressed) you get:
```
FileNotFoundError: Forced include not found: /private/var/folders/j_/kh9hgnt51jn_4rk3z3dz6m_h0000gr/T/pip-req-build-g5_xlihl/maya_submitter_plugin
```

Instead of packaging only the package for sdist, we also now include the custom script

buildable/installable sdist

```
hatch clean && hatch build && pip install --force-reinstall dist/*.tar.gz
```

N/A

N/A

### Does this change impact security?

N/A

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,8 @@ destinations = [
 
 [tool.hatch.build.targets.sdist]
 include = [
-    "src/*",
+    "src/deadline",
+    "maya_submitter_plugin",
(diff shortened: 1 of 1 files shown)


AWS-Samuel (approved):

jairaws:
Thank you!

jairaws (approved):

PR #287, opened 2025-06-13
fix: Maya scene with many asset files hangs without user feedback

leongdl:
Fixes: Customer reported issue. Maya submitter "hangs" when launching with thousands of assets. The 

- Customers have thousands of assets in Maya, and customers need to submit jobs using Maya to Deadline Cloud
- The problem happens because the dialog is harvesting all assets in the scene. It is done in a tight for loop that blocks Maya with no signal to the user.

- Allow time for the Qt event loop to run by adding appropriate sleeps and Maya UI refreshes.
- I selected to add interactivity rather than using threads because via research it seems Maya APIs are not very thread safe and needs to be run on the main thread.
- I added a progress dialog, along with signals to show the progress of scene introspection.

###
- Long term, I want to re-write this so the dialog is asynchronous from data harvesting. However, this needs us to re-write the dialog so it can fully support background operations and harvesting assets with the dialog open. On submission then it can block until the harvesting is fully done.

- However, that brings in a lot of other issues such as what to do with the Job Attachments tab while the harvesting is going on.

- Make sure the app is still interactive, and users know the system is running.

- hatch build && hatch run install
- Launch Maya, enable the deadline submitter.
- Loaded a scene with 14000 assets. Launch the submitter. See that we have a nice progress dialog now.

#### Unit test result
```
================================================================================================================================ slowest 5 durations =================================================================================================================================
0.12s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_start::test_maya_init_timeout
0.03s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_start::test_server_init_fail
0.03s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_run::test_run_data_wrong_schema
0.03s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_cleanup::test_on_cleanup
================================================================================================================================ 106 passed in 3.01s =================================================================================================================================
```

#### Integ test result
- hatch run integ:test_adaptors_maya
- I did not run redshift because I don't have redshift.
```
[gw0] [ 50%] PASSED test/integ/test_maya_adaptors.py::TestAdaptors::test_minimal_scene_adaptor 
test/integ/test_maya_submitters.py::TestSubmitters::test_minimal_scene_submitter 
[gw0] [100%] PASSED test/integ/test_maya_submitters.py::TestSubmitters::test_minimal_scene_submitter 
===Flaky Test Report===
```
#### Installer test result
- Not Applicable.

- Not Applicable

- Not Applicable

No, this is a submitter dialog gui change.

----

### DIFF
--- a/src/deadline/maya_submitter/assets.py
+++ b/src/deadline/maya_submitter/assets.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import os
 import re
+import time
 from functools import lru_cache
 from pathlib import Path
 from typing import Generator, Iterable
@@ -15,9 +16,12 @@ _FRAME_RE = re.compile("#+")
 
 
 class AssetIntrospector:
-    def parse_scene_assets(self) -> set[Path]:
+    def parse_scene_assets(self, progress_callback=None) -> set[Path]:
         """
         Searches the scene for assets, and filters out assets that are not needed for Rendering.
+
+        Args:
+            progress_callback: Optional callback function that takes a string argument for progress updates
 
         Returns:
             set[Path]: A set containing filepaths of assets needed for Rendering
@@ -28,14 +32,26 @@ class AssetIntrospector:
         assets: set[Path] = set()
 
         # Grab any yeti files
-        assets.update(self._get_yeti_files())
+        if progress_callback:
+            progress_callback("Searching for Yeti cache files...")
+        assets.update(self._get_yeti_files(progress_callback))
 
         if Scene.renderer() == RendererNames.arnold.value:
-            assets.update(self._get_tx_files())
+            if progress_callback:
+                progress_callback("Searching for Arnold texture files...")
+            assets.update(self._get_tx_files(progress_callback))
         elif Scene.renderer() == RendererNames.renderman.value:
-            assets.update(self._get_tex_files())
+            if progress_callback:
+                progress_callback("Searching for Renderman texture files...")
+            assets.update(self._get_tex_files(progress_callback))
 
-        for ref in FilePathEditor.fileRefs():
+        file_refs = list(FilePathEditor.fileRefs())
+        total_refs = len(file_refs)
+        print(f"Processing {total_refs} file references")
+        if progress_callback:
+            progress_callback(f"Processing {total_refs} file references...")
+
+        for i, ref in enumerate(file_refs):
             normalized_path = os.path.normpath(ref.path)
             # Files without tokens may already have been checked, if so, skip
             if normalized_path in assets:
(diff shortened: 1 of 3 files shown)


src/deadline/maya_submitter/assets.py
```diff
@@ -28,14 +33,26 @@ def parse_scene_assets(self) -> set[Path]:
...
         elif Scene.renderer() == RendererNames.renderman.value:
-            assets.update(self._get_tex_files())
+            if progress_callback:
+                progress_callback("Searching for Renderman texture files...")
+            assets.update(self._get_tex_files(progress_callback))
 
-        for ref in FilePathEditor.fileRefs():
+        file_refs = list(FilePathEditor.fileRefs())
+        total_refs = len(file_refs)
+        print(f"Processing {total_refs} file references")
```
jairaws, line 50: Q: Why are we using `print` instead of `log`? Where are these logs available? Is in the Maya scripting console?

src/deadline/maya_submitter/assets.py
```diff
@@ -45,29 +62,62 @@ def parse_scene_assets(self) -> set[Path]:
             # these files since it returns the original generator which is exhausted.
             for path in self._expand_path(normalized_path):
                 assets.add(path)
+            # Only refresh UI every 100 elements to improve performance
+            if i % 100 == 0:
```
jairaws, line 65: Not a blocker, but I feel it would be better to have a class to handle the progress logic. All the related logic feels a spread out.

src/deadline/maya_submitter/assets.py
```diff
@@ -3,22 +3,27 @@
 
 import os
 import re
+import time
 from functools import lru_cache
 from pathlib import Path
-from typing import Generator, Iterable
+from typing import Generator, Iterable, Optional, Callable
```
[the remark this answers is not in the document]
jairaws: We should address this to make the automatic validations pass.

src/deadline/maya_submitter/assets.py
leongdl: Only prints work from the maya console and general test output. Log doesn't print any user feedback to Maya.

src/deadline/maya_submitter/assets.py
leongdl: I was thinking about this, however the speed of the data harvest varies by code path so it is hard to "tune" this number. Empirically on my computer 100 was still interactive.

src/deadline/maya_submitter/assets.py
leongdl: Long term, I want to re-write this so the dialog is asynchronous from data harvesting. However, this needs us to re-write the dialog so it can fully support background operations and harvesting assets with the dialog open. On submission then it can block until the harvesting is fully done.

However, that brings in alot of other issues such as what to do with the Job Attachments tab while the harvesting is going on.

src/deadline/maya_submitter/assets.py
```diff
@@ -28,14 +32,26 @@ def parse_scene_assets(self) -> set[Path]:
...
+        if progress_callback:
+            progress_callback("Searching for Yeti cache files...")
+        assets.update(self._get_yeti_files(progress_callback))
 
         if Scene.renderer() == RendererNames.arnold.value:
-            assets.update(self._get_tx_files())
+            if progress_callback:
+                progress_callback("Searching for Arnold texture files...")
+            assets.update(self._get_tx_files(progress_callback))
         elif Scene.renderer() == RendererNames.renderman.value:
```
sakshie95, line 43: Woah we need to do this for every Renderer? How do we make sure it isn't missed for a future renderer we add?

src/deadline/maya_submitter/assets.py
```diff
@@ -112,13 +195,33 @@ def _get_tx_files(self) -> set[Path]:
...
+
+        for i, img_path in enumerate(texture_files):
             for expanded_path in self._expand_path(img_path):
                 arnold_textures_files.add(expanded_path)
                 # expanded files are guaranteed to exist, but we haven't checked the associated .tx file yet
                 if os.path.isfile(expanded_path.with_suffix(".tx")):
                     arnold_textures_files.add(expanded_path.with_suffix(".tx"))
 
+            # For every 100 texture files, update progress
+            if i % 100 == 0 and i > 0:
```
sakshie95, line 212: nit: This logic seems like it's repeated for every renderer? maybe a good idea to export these out to a common place?

test/unit/deadline_submitter_for_maya/scripts/test_assets.py
```diff
@@ -146,10 +146,8 @@ def test_get_tex_files(
 
     # python 3.9 3.10 requires to mock the import of maya.cmds
```
sakshie95, line 147: no unit tests for this dialog? its probably a suggestion for an integ test but related to my comment above: Maybe something that validates that this is implemented and we get a progress dialog for every supported renderer?

src/deadline/maya_submitter/assets.py
leongdl: Depends on the speed of the renderer to get tex files. Hard to say until we add them and test out the speed.

test/unit/deadline_submitter_for_maya/scripts/test_assets.py
leongdl: We can't add an integ test since its a UI thing. The submitter only verifies the job template works. UI tests are what I show'ed you.

I think Rena was working on submitter UI tests, but that's a different track.

src/deadline/maya_submitter/assets.py
sakshie95: I see.. might be a thing to consider during the release of every new renderer then. @benl-2023 this might be of something of note for you

sakshie95 (approved):

src/deadline/maya_submitter/assets.py
leongdl: Yes, i used 100 empirically, but other use cases maybe slower and 100 is not a good solution.

benl-2023 (approved):

jairaws (approved):

PR #292, opened 2025-06-19
test: Redshift adaptor & submitter integ test

benl-2023:
* We want to have an adaptor integration test for the redshift renderer in Maya. This will assert the correct behaviour of the redshift for maya adaptor handler if/when future modifications are made.
* We also want a new submitter test to validate the generate job bundle.

* Using the existing adaptor integration test pattern, a new scene and template were added that use redshift as the renderer. Output images are compared against the expected image in this PR.
* Modify the existing submitter integration test pattern to be parameterized with new test configurations

* Improved test coverage for adaptors and submitters

N/A See, below

#### Unit test result
N/A
#### Integ test result
```
test/integ/test_maya_adaptors.py::TestAdaptors::test_minimal_scene_adaptor 
[gw0] [ 25%] PASSED test/integ/test_maya_adaptors.py::TestAdaptors::test_minimal_scene_adaptor 
test/integ/test_maya_adaptors.py::TestAdaptors::test_redshift_scene_adaptor 
[gw0] [ 50%] PASSED test/integ/test_maya_adaptors.py::TestAdaptors::test_redshift_scene_adaptor 
test/integ/test_maya_submitters.py::TestSubmitters::test_scene_submitter[job_configuration0] 
[gw0] [ 75%] PASSED test/integ/test_maya_submitters.py::TestSubmitters::test_scene_submitter[job_configuration0] 
test/integ/test_maya_submitters.py::TestSubmitters::test_scene_submitter[job_configuration1] 
[gw0] [100%] PASSED test/integ/test_maya_submitters.py::TestSubmitters::test_scene_submitter[job_configuration1] 
```
#### Installer test result
N/A

No, this PR is only adding new test code.

N/A

No

----


test/integ/test_maya_adaptors.py
```diff
@@ -46,3 +46,26 @@ def test_minimal_scene_adaptor(self, script_location: Path, tmp_path: Path) -> N
             actual_image_directory=output_path,
             tolerance=2,
         )
+
+    def test_redshift_scene_adaptor(self, script_location: Path, tmp_path: Path) -> None:
```
leongdl, line 52: Can I run this myself on a setup w/o Redshift? Would it fail?

test/integ/test_maya_submitters.py
```diff
@@ -30,8 +31,10 @@ def initialize_maya():
         on_create_job_bundle_callback,
     )
 
+    qt_application = QtWidgets.QApplication(sys.argv)
```
leongdl, line 34: Why do we need to start the Qt event loop now?

test/integ/test_maya_adaptors.py
benl-2023: It would fail. That caveat is captured in the PR template.

The same will happen for the regular Maya adaptor test if you don't have Maya installed.

test/integ/test_maya_submitters.py
AlexTranAmz: I believe this is to ensure we only have 1 Qt session open per integ test run. As if we move it to run inside each submitters that it will crash due to multiple Qt session open

AlexTranAmz (dismissed):

test/integ/test_maya_submitters.py
leongdl: I see, because nested Qt event loops are forbidden (last time I used Qt) - should be OK as long as we do not end up with nested event loops.

test/integ/test_maya_adaptors.py
leongdl: Does it mean all of us need to setup redshift? Including contributors?

test/integ/test_maya_submitters.py
benl-2023: We always had to start the Qt event loop. Previously it was in the scope of the test case, but since we are running multiple test cases in this session, we need to move the Qt event loop into a session level fixture. 

Otherwise we will have issues with concurrent Qt applications running

test/integ/test_maya_adaptors.py
benl-2023: It will fail

AlexTranAmz: The idea is contributors should at least manually test the change, which mean they have Maya setup for the minimal integ test. We've also call out in this PR that the rest of the test here only need to pass if you make a modification to the corresponding adaptor, which go back to the first point of if they make the change, they should manually test it hence it should have been installed in that case.

hatch.toml
```diff
@@ -42,9 +42,11 @@ pre-install-commands = [
 ]
 
 [envs.integ.scripts]
-test = "mayapy -m pytest --no-cov {args:test/integ} -vvv --numprocesses=1"
-test_submitters = "mayapy -m pytest --no-cov {args:test/integ} -vvv --numprocesses=1 -m submitter"
-test_adaptors = "mayapy -m pytest --no-cov {args:test/integ} -vvv --numprocesses=1 -m adaptor"
+test = 'mayapy -m pytest --no-cov {args:test/integ} -vvv --numprocesses=1'
```
leongdl, line 45: nice! I like choices

leongdl (approved):

AlexTranAmz (approved):

PR #295, opened 2025-06-26
feat: Added validation for frame range field in submitter ui

kenyrish-amazon:
Fixes: Added validation for the override frame range value field in submitter UI

The Override frame range field in the submitter accepts all characters, but only numbers, whitespace, and other characters to create a range should be allowed ("-" "," ":")

Created a regex validation for the text box so characters other than the ones specified cannot be entered.

Invalid characters cannot be entered in the text box. Invalid frame ranges can still be created with just these characters, and the job will still fail.

Manually tested, made sure the regex works and only the specified characters are allowed. Also created a test case to verify that the validator is set. Used similar test case to [here](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/e7d62dcb6c0226c4f056ac84835fbef406691683/test/unit/deadline_submitter_for_cinema4d/test_submitter.py#L34)

#### Unit test result
```
================================================================================== slowest 5 durations ==================================================================================
0.12s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_start::test_maya_init_timeout
0.03s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_run::test_run_data_wrong_schema
0.03s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_start::test_no_error
0.03s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_start::test_populate_action_queue_test_mapping
================================================================================== 107 passed in 2.88s ==================================================================================
```

No

No

### DIFF
--- a/src/deadline/maya_submitter/ui/components/scene_settings_tab.py
+++ b/src/deadline/maya_submitter/ui/components/scene_settings_tab.py
@@ -1,7 +1,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 import os
 
-from qtpy.QtCore import QSize, Qt  # type: ignore
+from qtpy.QtCore import QSize, Qt, QRegularExpression  # type: ignore
 from qtpy.QtWidgets import (  # type: ignore
     QCheckBox,
     QComboBox,
@@ -15,6 +15,7 @@ from qtpy.QtWidgets import (  # type: ignore
     QSpacerItem,
     QWidget,
 )
+from qtpy.QtGui import QRegularExpressionValidator  # type: ignore
 from deadline.client.ui import block_signals
 
 from ...render_layers import LayerSelection
@@ -125,6 +126,9 @@ class SceneSettingsWidget(QWidget):
 
         self.frame_override_chck = QCheckBox("Override Frame Range", self)
         self.frame_override_txt = QLineEdit(self)
+        # Only allow numbers, colons, dashes, commas, and whitespace for frame ranges
+        frame_pattern = QRegularExpression(r"^[0-9:\-,\s]*$")
+        self.frame_override_txt.setValidator(QRegularExpressionValidator(frame_pattern))
         lyt.addWidget(self.frame_override_chck, 4, 0)
         lyt.addWidget(self.frame_override_txt, 4, 1)
         self.frame_override_chck.stateChanged.connect(self.activate_frame_override_changed)
--- a/test/unit/deadline_submitter_for_maya/scripts/test_submitter.py
+++ b/test/unit/deadline_submitter_for_maya/scripts/test_submitter.py
@@ -1,0 +1,43 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import sys
+from unittest.mock import Mock
+
+
+def test_frame_override_has_text_validation():
+    try:
+        # Create new mocks. We need to use a real class for QWidget instead of a Mock() so that the
+        # SceneSettingsWidget which is a subclass of QWidget actually runs its methods (instead its
+        # methods being mocked out).
+        class MockQWidget:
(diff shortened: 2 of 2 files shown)


src/deadline/maya_submitter/ui/components/scene_settings_tab.py
```diff
@@ -125,6 +126,9 @@ def _build_ui(self, settings):
 
         self.frame_override_chck = QCheckBox("Override Frame Range", self)
         self.frame_override_txt = QLineEdit(self)
+        # Only allow numbers, colons, dashes, commas, and whitespace for frame ranges
+        frame_pattern = QRegularExpression(r"^[0-9:\-,\s]*$")
```
godobyte, line 130: Is it a legitimate use case to have float for frame ranges? I remember hearing about it somewhere, could you double check for Maya?

kenyrish-amazon: I was looking at [this](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#34111-intrangeexpr) for the specifications, which says the frame range only handles ints. Also, if I put a float in the range and submit, it causes an error. Is there another place that has the specifications for Maya?

mwiebe: When we looked at frame ranges for OpenJD, we went partly down the path of specifying float range expressions too, but stuck with just integers in the first revision to focus time on other parts of the spec. Floats are harder to spell out precisely, and I don't recall seeing it done in other DCCs the way Maya does it.

What do your use cases look like for this? I think we could handle it in Maya specifically by multiplying the frames by 10 or 100, and then having the adaptor part divide it back out. E.g. Something like "0-3.5:0.5" would be "0-35:5" and every frame number would be divided by 10 before going to Maya.

kenyrish-amazon: The [frame range class ](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/9b142a54728ffcb363e78e352c89f1e3098d8e1a/src/deadline/maya_submitter/scene.py#L194) expects ints for the start, stop and step variables, so for this validation I think it makes sense to not allow floats to be entered. 
Your suggestion makes sense, but supporting floats would be a separate feature, and I'm not aware of a current use case that would require it.

test/unit/deadline_submitter_for_maya/scripts/test_submitter.py
```diff
@@ -0,0 +1,43 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import sys
+from unittest.mock import Mock
+
+
+def test_frame_override_has_text_validation():
```
benl-2023, line 7: This looks like a good test to verify we loaded a validator into the UI. Is it possible as a unit test to also test out submissions of good and bad frame ranges?

benl-2023 (approved):

test/unit/deadline_submitter_for_maya/scripts/test_submitter.py
AlexTranAmz: I think this is more of integ/e2e test due to the nature of trying to submit a job and need to run all the open-jd template validation

kenyrish-amazon: Discussed this, it is hard to test these submissions since QWidget is being mocked automatically so the regex value is not stored here.

AlexTranAmz (approved):

PR #308, opened 2025-08-15
test: add unit tests for submitter assets.py

jericht:
While trying to add unit test cases for the changes in https://github.com/aws-deadline/deadline-cloud-for-maya/pull/259, I discovered that there is no existing unit tests for the submitter's `assets.py` file.

Rather than adding all the unit tests in the above PR, I am backfilling unit test coverage for existing code in this separate PR.

~~I also found some unused functions in a utils.py file, I've removed the file and relocated the only used function to the `assets.py` file, the only place it is used.~~

**Note**: The original code in the `test_assets.py` file did not look like it was actually testing what it claimed to test. There was also an unused test class in the file. It felt like an abandoned file, so I removed all the existing contents (2 tests) and wrote the tests from scratch. When reviewing, I recommend just viewing the new contents.

We have unit tests for submitter's `assets.py` file.

Ran the unit tests `hatch run test`

#### Unit test result

```
Required test coverage of 41.0% reached. Total coverage: 48.68%
=============================================================================================================================== slowest 5 durations ================================================================================================================================
0.13s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_start::test_maya_init_timeout
0.03s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_stop::test_on_stop
0.03s call     test/unit/deadline_submitter_for_maya/scripts/test_assets.py::test_get_tex_files
0.03s call     test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py::TestMayaAdaptor_on_start::test_no_error
=============================================================================================================================== 112 passed in 2.40s ================================================================================================================================
```

#### Integ test result

SKIPPED - this only affects unit tests

#### Installer test result

SKIPPED - this only affects unit tests

SKIPPED - this only affects unit tests

No, it is unit tests

No

----


src/deadline/maya_submitter/assets.py
```diff
@@ -284,7 +285,41 @@ def _expand_path(self, path: str) -> Generator[Path, None, None]:
...
             for group in frame_re_matches:
                 working_path = working_path.replace(group, str(frame).zfill(len(group)))
-            paths = findAllFilesForPattern(working_path, frame)
+            paths = _findAllFilesForPattern(working_path, frame)
             for p in paths:
                 if not p.endswith(":Zone.Identifier"):  # Metadata files that erroneously match
                     yield Path(p)
+
+
+def _findAllFilesForPattern(pattern: str, frameNumber: int) -> list[str]:
```
leongdl: Any specific reason to delete `utils.py` ?

leongdl (dismissed):

src/deadline/maya_submitter/assets.py
jericht: Good catch, that would be a breaking change. Restored `utils.py` to its original state.

leongdl (approved):
Thanks for keeping the methods for backwards compatibility.

jairaws (approved):

PR #317, opened 2025-09-02
test: add unit tests for cache file scene introspection

jericht:
In https://github.com/aws-deadline/deadline-cloud-for-maya/pull/259, we improved the scene introspection to include  files that are referenced via attributes on nodes (e.g. Bifrost cache files), however we decided to go ahead and merge and backfill unit tests for this functionality later. We should backfill the unit tests for these changes.

Add the unit tests for https://github.com/aws-deadline/deadline-cloud-for-maya/pull/259

We have unit tests for https://github.com/aws-deadline/deadline-cloud-for-maya/pull/259

`hatch run test`

#### Integ test result

N/A

#### Installer test result
N/A

No, this only adds unit tests

No

No

----

### DIFF
--- a/test/unit/deadline_submitter_for_maya/scripts/test_utils.py
+++ b/test/unit/deadline_submitter_for_maya/scripts/test_utils.py
@@ -64,6 +64,7 @@ def test_join_paths(first_path: str, second_path: str, expected_output: str):
     assert join_paths(first_path, second_path) == expected_output
 
 
+@pytest.mark.parametrize("frame_symbol", ["<f>", "<frame>"])
 @patch.object(utils_module.os.path, "isfile", return_value=True)
 @patch.object(utils_module.os, "listdir")
 @patch.object(utils_module.os.path, "isdir", return_value=True)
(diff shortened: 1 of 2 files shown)


test/unit/deadline_submitter_for_maya/scripts/test_utils.py
```diff
@@ -85,7 +87,9 @@ def test_findAllFilesForPattern(
         "/test/file/path.2.png",
     ]
     # TODO: Figure out what patternToRegex actually does
```
godobyte, line 89: nit - is this still applicable?

godobyte (approved):

sakshie95 (approved):

test/unit/deadline_submitter_for_maya/scripts/test_utils.py
jericht: Yeah.. I couldn't find any info/docs on it, so the test assumes it just does this for now. Either way, test is fine.. just would be nicer if it was being mocked closer to what the actual values are.

PR #343, opened 2025-10-24
feat: Add V-Ray tile rendering

jblagden:
### What was the problem/requirement?

The Maya adaptor cannot output tiled renders from V-Ray.

### What was the solution?

We can instruct the V-Ray frame buffer to only render a sub-set of the final image.

We can get tiled output as EXRs from a V-Ray render.

This relies on the job bundle in [PR #142 on the deadline-cloud-samples repository](https://github.com/aws-deadline/deadline-cloud-samples/pull/142).

The testing steps are the same.

#### Integ test result

*delete text starting here*
If `src/` was modified or a file was added/removed, then update the integ tests and post the test results below. See
DEVELOPMENT.md on additional instructions on running some, or all integ tests.
*delete text ending here*

#### Installer test result

Installer was not touched.

No, all changes are under the V-Ray handler and the integration tests use Redshift.

Yes

No

----

### DIFF
--- a/src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
+++ b/src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
@@ -46,8 +46,13 @@ class VRayHandler(DefaultMayaHandler):
             data (dict): The data given from the Adaptor. Keys expected: ['frame']
 
         Raises:
-            RuntimeError: If no camera was specified or no renderable camera was found,
-            when no frame was specified or when no vraySettings node found in the scene.
+            RuntimeError: Thrown when a render-time error is found.
+            Causes include:
+            - If no camera was specified
+            - No renderable camera was found
+            - No frame was specified
+            - No vraySettings node found in the scene
+            - Output image type is not exr for a region render
         """
         if not maya.cmds.pluginInfo("vrayformaya", query=True, loaded=True):
             raise RuntimeError(
@@ -63,6 +68,11 @@ class VRayHandler(DefaultMayaHandler):
             raise RuntimeError("MayaClient: start_render called without a valid camera.")
         self.render_kwargs["camera"] = self.camera_name
 
+        if not self.vraySettingsNodeExists():
+            raise RuntimeError(
+                "MayaClient: start_render called with missing vraySettings node in the scene."
+            )
+
         # In order of preference, use the task's output_file_prefix, the step's output_file_prefix, or the scene file setting.
         output_file_prefix = data.get("output_file_prefix", self.output_file_prefix)
         if output_file_prefix:
@@ -76,20 +86,6 @@ class VRayHandler(DefaultMayaHandler):
         if self.image_height is not None:
             maya.cmds.setAttr("vraySettings.height", self.image_height)
             print(f"Set image height to {self.image_height}", flush=True)
-
-        region = [
-            data.get(field)
-            for field in ("region_min_x", "region_max_x", "region_min_y", "region_max_y")
-        ]
-        if any(v is not None for v in region):
-            raise RuntimeError(
-                "MayaClient: A region render was specified, but region rendering support is not implemented for the selected renderer."
-            )
-
-        if not self.vraySettingsNodeExists():
-            raise RuntimeError(
-                "MayaClient: start_render called with missing vraySettings node in the scene."
-            )
 
         # Always set the animation type to 2 (specific frames) and one frame animation
         maya.cmds.setAttr("vraySettings.animType", 2)
@@ -115,6 +111,54 @@ class VRayHandler(DefaultMayaHandler):
         # Set the log message level to 3 (report errors, warnings and general information) if needed
         if maya.cmds.getAttr("vraySettings.sys_message_level") < 3:
             maya.cmds.setAttr("vraySettings.sys_message_level", 3)
+
+        # Perform setup for region rendering if needed, otherwise just use the output size from the submission
+        region = [
+            data.get(field)
+            for field in ("region_min_x", "region_min_y", "region_max_x", "region_max_y")
+        ]
+        if any(v is not None for v in region):
+            print(f"MayaClient: Region bounds {region} specified.", flush=True)
+
+            region_minX, region_minY, region_maxX, region_maxY = region
+            region_str = (
+                f"(minX={region_minX}, minY={region_minY}, maxX={region_maxX}, maxY={region_maxY})"
+            )
+            if any(v is None for v in region):
+                raise RuntimeError(
+                    f"MayaClient: Region bounds {region_str} must be fully defined or all empty, but were partially specified."
+                )
+
+            # Set the output filename
+            maya.cmds.setAttr(
+                "vraySettings.fileNamePrefix", data.get("output_file_prefix"), type="string"
+            )
+
+            # Set to allow region in batch rendering
+            maya.cmds.setAttr("vraySettings.vfbRgnOffBatch", 0)
+
+            # Set to use VFB
+            maya.cmds.setAttr("vraySettings.vfbOn", 1)
+
+            # Make sure the image output is set to EXR. If it isn't the region render will not work so we must fail
+            ouput_extension = maya.cmds.getAttr("vraySettings.imageFormatStr")
+
+            # The exr extension name can be one of "exr", "exr (deep)", or "exr (multichannel)"
+            if "exr" not in ouput_extension.lower():
+                raise RuntimeError(
+                    f'MayaClient: Output image format is set to "{ouput_extension}" but must be EXR for region rendering to work.'
+                )
+
+            # Set the region to render
+            region_cmd = f"vray vfbControl -setregion {region_minX} {region_minY} {region_maxX} {region_maxY}; vraySetBatchDoRegion {region_minX} {region_minY} {region_maxX} {region_maxY};"
+
+            print(f"Setting render region: {region_cmd}")
+            maya.mel.eval(region_cmd)
+
+            get_region_cmd = "vray vfbControl -getregion"
+            print(
+                f"Checking the region is set correctly: {maya.mel.eval(get_region_cmd)}", flush=True
+            )
 
         maya.cmds.vrend(**self.render_kwargs)
         print(f"MayaClient: Finished Rendering Frame {frame}\n", flush=True)


src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
```diff
@@ -116,6 +107,43 @@ def start_render(self, data: dict) -> None:
...
+
+            region_minX, region_minY, region_maxX, region_maxY = region
+            region_str = f"(minX={region_minX}, minY={region_minY}, maxX={region_maxX}, maxY={region_maxY})"
+            if any(v is None for v in region):
+                raise RuntimeError(
+                    f"MayaClient: Region bounds {region_str} must be fully defined or all empty, but were partially specified."
+                )
+
+            # Set the output filename
+            maya.mel.eval(f'''setAttr -type "string" "vraySettings.fileNamePrefix" "{data.get("output_file_prefix")}";''')
```
mwiebe: Can you do this with a `maya.cmds.setAttr` call instead of an `eval` to a setAttr command?

src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
```diff
@@ -116,6 +107,43 @@ def start_render(self, data: dict) -> None:
...
+
+            # Set to allow region in batch rendering
+            maya.cmds.setAttr("vraySettings.vfbRgnOffBatch", 0)
+
+            # Set to use VFB
+            maya.cmds.setAttr("vraySettings.vfbOn", 1)
+
+            # Set the image format to EXR - This is needed to make sure image is properly cropped
```
mwiebe: It should probably raise an error if the image format is different from EXR, instead of overriding the format.

mwiebe (changes requested):

src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
jblagden: I had tried doing that, but when I did the command didn't apply any changes.

As far as I can tell it should work the same, but the vraysettings node wasn't being updated.

src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
jblagden: Good call, added that and tested with a png output and exr output job.

src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
mwiebe: Got it. Can you update the comment to describe that we found calling `maya.cmds.setAttr` did not work?

mwiebe: Another question then, I think `"{data.get("output_file_prefix")}"` could fail if the value contains special characters. Maybe we could use `shlex` to quote it correctly?

jblagden: I'm not sure I follow you on where `shlex` would be useful here?

Something like this?
```
safe_prefix = shlex.quote(data.get("output_file_prefix"))
maya.mel.eval(f'''setAttr -type "string" "vraySettings.fileNamePrefix" "{safe_prefix}";''')
```

I don't know if we can assume Maya's quoting is going to line up with what a [Unix shell expects and shelx.quote outputs](https://docs.python.org/3/library/shlex.html#shlex.quote) but I'll see if I can break my current implementation and see how it does. If I can break it, I can prove this fixes it.

I'll also grab some Maya + Vray version numbers to explain why we're not just calling
```
maya.cmd.setAttr(vraySettings.FileNamePrefix, data.get("output_file_prefix"))
```

for that comment as those are missing from my notes.

mwiebe: Maybe `json.dumps` would produce a compatible escaped string?

mwiebe: Here's some examples of trying it:

```python
>>> output_file_prefix = "value with \" and ' and \U0000263a"
>>> print(output_file_prefix)
value with " and ' and ☺
>>> import shlex, json

>>> print(shlex.quote(output_file_prefix))
'value with " and '"'"' and ☺'
>>> print(json.dumps(output_file_prefix))
"value with \" and ' and \u263a"

>>> print(f'''setAttr -type "string" "vraySettings.fileNamePrefix" {json.dumps(output_file_prefix)}''')
setAttr -type "string" "vraySettings.fileNamePrefix" "value with \" and ' and \u263a"
```

jblagden: I've just tried using `value with \" and ' and \U0000263a` as the output file prefix, and it's the adaptor that dies trying to parse it as a part of the YAML.

```
2025/11/03 14:09:51-06:00 Command started as pid: 8653
2025/11/03 14:09:51-06:00 Output:
2025/11/03 14:09:51-06:00 frame: 1
2025/11/03 14:09:51-06:00 region_min_x: 0
2025/11/03 14:09:51-06:00 region_max_x: 50
2025/11/03 14:09:51-06:00 region_min_y: 0
2025/11/03 14:09:51-06:00 region_max_y: 50
2025/11/03 14:09:51-06:00 output_file_prefix: '<Scene>value with \" and ' and \U0000263a_tile_1x1_2x2'
2025/11/03 14:09:51-06:00 Failed to load data as JSON or YAML: while parsing a block mapping
2025/11/03 14:09:51-06:00   in "run-data.yaml", line 1, column 1
2025/11/03 14:09:51-06:00 expected <block end>, but found '<scalar>'
2025/11/03 14:09:51-06:00   in "run-data.yaml", line 6, column 49
2025/11/03 14:09:51-06:00 Error parsing command line arguments: while parsing a block mapping
2025/11/03 14:09:51-06:00   in "run-data.yaml", line 1, column 1
2025/11/03 14:09:51-06:00 expected <block end>, but found '<scalar>'
2025/11/03 14:09:51-06:00   in "run-data.yaml", line 6, column 49
2025/11/03 14:09:51-06:00 Entrypoint failed: while parsing a block mapping
2025/11/03 14:09:51-06:00   in "run-data.yaml", line 1, column 1
2025/11/03 14:09:51-06:00 expected <block end>, but found '<scalar>'
2025/11/03 14:09:51-06:00   in "run-data.yaml", line 6, column 49
2025/11/03 14:09:51-06:00 Traceback (most recent call last):
2025/11/03 14:09:51-06:00   File "/sessions/session-bc6af46020f9477a8082ad6fb227229f43mjyq78/embedded_filespi3poahg/render-tile.py", line 36, in <module>
2025/11/03 14:09:51-06:00     subprocess.check_call(["maya-openjd", "daemon", "run", "--run-data", "file://run-data.yaml"])
2025/11/03 14:09:51-06:00   File "/usr/lib64/python3.9/subprocess.py", line 373, in check_call
2025/11/03 14:09:51-06:00     raise CalledProcessError(retcode, cmd)
2025/11/03 14:09:51-06:00 subprocess.CalledProcessError: Command '['maya-openjd', 'daemon', 'run', '--run-data', 'file://run-data.yaml']' returned non-zero exit status 1.
2025/11/03 14:09:51-06:00 Process pid 8653 exited with code: 1 (unsigned) / 0x1 (hex)
```

I'm going to carry on testing. It looks like we need to be careful about special characters further up the stack; before Maya gets them at least.

jblagden: Using `maya.cmds.setAttr("vraySettings.fileNamePrefix", data.get("output_file_prefix"), type="string")` just works, and I've got it committed.

This is using the maya-vray-2025.62002 and maya-2025.0 packages from the deadline-cloud channel.

mwiebe (dismissed):

src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
```diff
@@ -116,6 +112,61 @@
...
+            # Make sure the image output is set to EXR. If it isn't the region render will not work so we must fail
+            ouput_extension = maya.cmds.getAttr("vraySettings.imageFormatStr")
+
+            # The exr extension name can be one of "exr", "exr (deep)", or "exr (multichannel)"
+            if "exr" not in ouput_extension.lower():
+                raise RuntimeError(
+                    f'MayaClient: Output image format is set to "{ouput_extension}" but must be EXR for region rendering to work.'
+                )
+
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
jblagden: I can't believe I've done this

CHANGELOG.md
```diff
@@ -1,24 +1,3 @@
-## 0.15.12 (2025-09-29)
```
mwiebe, line 1: I don't think you meant to change this file here.

mwiebe (approved):

leongdl (approved):
LGTM

src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py
```diff
@@ -116,6 +112,54 @@ def start_render(self, data: dict) -> None:
...
+            if "exr" not in ouput_extension.lower():
+                raise RuntimeError(
+                    f'MayaClient: Output image format is set to "{ouput_extension}" but must be EXR for region rendering to work.'
+                )
+
+            # Set the region to render
+            region_cmd = f"vray vfbControl -setregion {region_minX} {region_minY} {region_maxX} {region_maxY}; vraySetBatchDoRegion {region_minX} {region_minY} {region_maxX} {region_maxY};"
+
+            print(f"Setting render region: {region_cmd}")
+            maya.mel.eval(region_cmd)
```
leongdl, line 156: I forgot what our stance was to mix script and python together.

mwiebe: I try to avoid `mel.eval`, and only resort to it when the alternative is unavailable or not working. I don't know if this `vray` command is available via Python exposure. The command itself looks low risk to me, since the parameters are all integers there isn't any tricky string quoting needed.

PR #361, opened 2026-02-02
fix: enable custom ocio config for submitter

seant-aws:

Maya has OCIO config files for color management. Scenes can have custom OCIO configs, or default maya ones. When submitting jobs with custom OCIO configs, Job Attachments detects it properly, but the Maya adaptor does not properly set the custom OCIO when rendering the scene.

Fixed bug where path mapping was not being properly applied to OCIO configs, due to chicken and egg problem. To get the OCIO config path, we must first open the scene file. But opening the scene file automatically loads the unmapped OCIO config path, leading to error and automatically enabling the default Maya OCIO.

Added fix parses the OCIO path in the submitter, and passes to adaptor, which mapping is then applied and set to the maya instance.

Maya renders can now properly use custom OCIO config files.

ran `hatch` unit tests
ran `hatch` integ tests

### REPRODUCED ERROR:

output form deadline cloud monitor, under task output logs
setting a custom config path under `/Users/stangch/DeadlineClient/mayaTests/ocio_test/config.ocio` leads to this output:
```
2026/01/28 13:37:18-08:00 ADAPTOR_OUTPUT: STDERR: Error: Color management settings cannot be applied: Error could not read '/Users/stangch/DeadlineClient/mayaTests/ocio_test/config.ocio' OCIO profile..

2026/01/28 13:37:20-08:00 ADAPTOR_OUTPUT: STDOUT: 00:00:04   868MB         
| [color_manager] using color manager defaultColorMgtGlobals of type "color_manager_ocio"
2026/01/28 13:37:20-08:00 ADAPTOR_OUTPUT: STDOUT: 00:00:04   869MB        | [color_manager_ocio] defaultColorMgtGlobals : using OCIO configuration file /sessions/session-a09e882b274b430ab7cf66279714df04pl4z7q2z/.env/usr/autodesk/maya2026/resources/OCIO-configs/Maya2022-default/config.ocio
2026/01/28 13:37:20-08:00 ADAPTOR_OUTPUT: STDOUT: 00:00:04   870MB        | [color_manager] rendering color space is "ACEScg"
```
incorrectly using thedefault maya `ocio.config`

### AFTER FIX:
```
2026/02/02 13:04:34-08:00 ADAPTOR_OUTPUT: STDOUT: Setting OCIO config: '/sessions/session-c42a9eca3e654a699ffb9178fb6ecc20zmo8c_m5/assetroot-f55c5bc9cc6737a83d5e/DeadlineClient/mayaTests/ocio_test/jericht_custom_config.ocio'


2026/02/02 13:04:36-08:00 ADAPTOR_OUTPUT: STDOUT: 00:00:04   907MB         | [color_manager] using color manager defaultColorMgtGlobals of type "color_manager_ocio"
2026/02/02 13:04:36-08:00 ADAPTOR_OUTPUT: STDOUT: 00:00:04   908MB         | [color_manager_ocio] defaultColorMgtGlobals : using OCIO configuration file /sessions/session-c42a9eca3e654a699ffb9178fb6ecc20zmo8c_m5/assetroot-f55c5bc9cc6737a83d5e/DeadlineClient/mayaTests/ocio_test/jericht_custom_config.ocio
2026/02/02 13:04:36-08:00 ADAPTOR_OUTPUT: STDOUT: 00:00:04   908MB         | [color_manager] narrow color space "sRGB" is linear, and will cause banding for narrow bit widths: use an sRGB equivalent
2026/02/02 13:04:36-08:00 ADAPTOR_OUTPUT: STDOUT: 00:00:04   908MB         | [color_manager] unable to determine rendering color space chromaticities, assuming "sRGB"
```

#### Integ test result

```
=================================== slowest 5 durations ===================================
3.98s call     test/integ/test_maya_submitters.py::TestSubmitters::test_scene_submitter[Redshift Test]
2.12s setup    test/integ/test_maya_submitters.py::TestSubmitters::test_scene_submitter[Minimal Maya Test]
1.45s call     test/integ/test_maya_submitters.py::TestSubmitters::test_scene_submitter[Minimal Maya Test]
0.06s teardown test/integ/test_maya_submitters.py::TestSubmitters::test_scene_submitter[Redshift Test]
0.00s setup    test/integ/test_maya_submitters.py::TestSubmitters::test_scene_submitter[Redshift Test]
```

modified integ tests to introduce optional OCIO config

Updated schema file along with adaptor change, will be separate PR.

- [x] Yes
- [ ] No

   - [x] Yes
   - [ ] No

no, OCIO config files are optional. There are default values provided to fallback on, since not all scenes have a custom OCIO.

----

### DIFF
--- a/src/deadline/maya_adaptor/MayaAdaptor/adaptor.py
+++ b/src/deadline/maya_adaptor/MayaAdaptor/adaptor.py
@@ -43,6 +43,7 @@ _MAYA_INIT_KEYS = {
     "camera",
     "image_height",
     "image_width",
+    "ocio_config_file",
     "output_file_path",
     "output_file_prefix",
     "render_layer",
@@ -89,7 +90,7 @@ class MayaAdaptor(Adaptor[AdaptorConfiguration]):
 
     @property
     def integration_data_interface_version(self) -> SemanticVersion:
-        return SemanticVersion(major=0, minor=1)
+        return SemanticVersion(major=0, minor=2)
 
     @staticmethod
     def _get_timer(timeout: int | float) -> Callable[[], bool]:
--- a/src/deadline/maya_adaptor/MayaAdaptor/schemas/init_data.schema.json
+++ b/src/deadline/maya_adaptor/MayaAdaptor/schemas/init_data.schema.json
@@ -6,6 +6,7 @@
         "error_on_arnold_license_fail": { "type": "boolean" },
         "image_height": { "type": "number" },
         "image_width": { "type": "number" },
+        "ocio_config_file": { "type": "string" },
         "output_file_path": { "type": "string" },
         "output_file_prefix": { "type": "string" },
         "project_path": { "type": "string" },
--- a/src/deadline/maya_submitter/scene.py
+++ b/src/deadline/maya_submitter/scene.py
@@ -189,6 +189,32 @@ class Scene:
                 files.append(cache_file)
         return files
 
+    @staticmethod
+    def ocio_config_file() -> Optional[str]:
+        """
+        Returns the OCIO config file path if color management is enabled
+        and using a custom OCIO configuration.
+
+        Returns:
+            Optional[str]: The absolute path to the OCIO config file,
+                          or None if using default color management or no custom config
+        """
+        # Check if color management is enabled
+        cm_enabled = maya.cmds.colorManagementPrefs(query=True, cmEnabled=True)
+        if not cm_enabled:
+            return None
+
+        # Get the config file path - Maya returns bool if no path set
+        config_path = maya.cmds.colorManagementPrefs(query=True, configFilePath=True)
+        if not isinstance(config_path, str):
+            return None
+
+        # Skip Maya default configs
+        if "<MAYA_RESOURCES>" in config_path:
+            return None
+
+        return config_path
+
 
 @dataclass
 class FrameRange:
--- a/test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py
+++ b/test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py
@@ -30,6 +30,7 @@ EXPECTED_INIT_DATA_PROPERTIES = {
     "error_on_arnold_license_fail": True,
     "image_height": 1080,
     "image_width": 1920,
+    "ocio_config_file": "/path/to/config.ocio",
     "output_file_path": "/path/to/output",
     "output_file_prefix": "<Scene>/<RenderLayer>",
     "project_path": "/path/to/project",
@@ -61,7 +62,7 @@ EXPECTED_RUN_DATA_REQUIRED = ["frame"]
 
 # Expected version - must be bumped when schemas change
 EXPECTED_SCHEMA_VERSION_MAJOR = 0
-EXPECTED_SCHEMA_VERSION_MINOR = 1
+EXPECTED_SCHEMA_VERSION_MINOR = 2
 
 
 @pytest.fixture()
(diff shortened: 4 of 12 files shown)


src/deadline/maya_submitter/scene.py
```diff
@@ -189,6 +189,32 @@ def yeti_cache_files() -> list[str]:
...
+        Returns:
+            Optional[str]: The absolute path to the OCIO config file,
+                          or None if using default color management or no custom config
+        """
+        # Check if color management is enabled
+        cm_enabled = maya.cmds.colorManagementPrefs(query=True, cmEnabled=True)
+        if not cm_enabled:
+            return None
+
+        # Get the config file path - Maya returns bool if no path set
```
godobyte, line 207: Curious about `Maya returns bool if no path set`, under what circumstances does it return true and false?

test/unit/deadline_adaptor_for_maya/MayaAdaptor/test_adaptor.py
```diff
@@ -61,7 +62,7 @@
 
 # Expected version - must be bumped when schemas change
 EXPECTED_SCHEMA_VERSION_MAJOR = 0
-EXPECTED_SCHEMA_VERSION_MINOR = 1
+EXPECTED_SCHEMA_VERSION_MINOR = 2
```
godobyte, line 65: Q - which schema, the `init_data.schema.json`, why is that?

seant-aws: oh this is because I updated the template for init data, including the optional `ocio_config` field, and there's some tests/docs in place to make sure that the schema version is bumped when theres a change. This is needed for passing data from submitter to adaptor, since I'm grabbing the path from the submitter side and path mapping it when its on the adaptor side

src/deadline/maya_submitter/scene.py
seant-aws: the maya api is a bit strange, it will return strings if there is an existing path for this config, but will return `true` if it queries and there is no path I believe. As for `false` im not sure if that will happen, but just in case. It theoretically won't happen

godobyte (dismissed):

src/deadline/maya_adaptor/MayaAdaptor/adaptor.py
jericht, line 1: I think we're missing corresponding changes in the `MayaClient` to handle the new `ocio_config_file`, no?

seant-aws: yea you're right, I was planning to do so in an adaptor side PR so that this PR wouldn't be so beefy

jericht (approved):

src/deadline/maya_adaptor/MayaAdaptor/schemas/init_data.schema.json
```diff
@@ -6,6 +6,7 @@
         "error_on_arnold_license_fail": { "type": "boolean" },
```
leongdl, line 6: Meta comment - since the ocio file is auto loaded on the worker anyways, why did we have to add so much stuff from the submitter all the way through.

Can we traverse the scene graph on the worker, and then just directly apply path mapping?

leongdl: Ok chatted offline - seems there is no way to intercept the default behavior so we can take action.

godobyte (approved):

src/deadline/maya_adaptor/MayaAdaptor/schemas/init_data.schema.json
seant-aws: talked about this in person, just to summarize:

because opening scene file with unmapped path causes it to discard and use default path, we must pass the `ocio.config` path from submitter to adaptor at submission time.

PR #362, opened 2026-02-03
docs: maya adaptor/submitter development.md clarity improvements

seant-aws:
During my Maya development ramp up, there were tiny details I had to clarify, which were helpful.

I added these comments to the documentation

Improved details when doing local development and testing

No testing, just rereading the docs

- [ ] Yes
- [x] No

   - [ ] Yes
   - [x] No

No

----

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -84,12 +84,15 @@ export DEADLINE_ENABLE_DEVELOPER_OPTIONS=true
 maya
 ```
 
-```bash
+```powershell
 # Windows
 hatch shell
 $env:DEADLINE_ENABLE_DEVELOPER_OPTIONS = "true"
 maya
 ```
+
+**Note:** The `hatch shell` environment sets `MAYA_ENV_DIR` to point to `plugin_env/`, which causes Maya to load
+the development version of the plug-in. You must launch Maya from within `hatch shell` for this to work.
 
 You will need to load the plug-in within Maya once the application has completed loading. In the main menu bar, go to
 Windows > Settings/Preferences > Plug-In Manager and you will find that `DeadlineCloudForMaya.py` is available as a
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -301,16 +304,18 @@ You'll need to perform the following steps to substitute your build of the adapt
...
+   ls ./wheels
    deadline_cloud_for_maya-<version>-py3-none-any.whl
    deadline-<version>-py3-none-any.whl
    openjd_adaptor_runtime-<version>-py3-none-any.whl
    ```
 
-3. Open the Maya integrated submitter, and in the Job-Specific Settings tab, enable the option 'Include Adaptor Wheels'. This option is only visible when the environment variable `DEADLINE_ENABLE_DEVELOPER_OPTIONS` is set to `true`. Then submit your test job.
+4. Open the Maya integrated submitter, and in the Job-Specific Settings tab, enable the option 'Include Adaptor Wheels'. This option is only visible when the environment variable `DEADLINE_ENABLE_DEVELOPER_OPTIONS` is set to `true`. Then submit your test job.
+
+**Note:** The `wheels/` generated folder need to be copied into `plugin_env/wheels/` after each `hatch run install` for the submitter plug-in dependencies. This ensures local adaptor changes are reflected on your local Maya instance launched in `hatch shell`.
```
jairaws, line 318: Could this be automated? Does not need to be done now, but it is not a good CX having to copy the folder manually.

jairaws (approved):

jairaws:
The title of the PR is a bit misleading. It reads like we are changing the process.

seant-aws:
> The title of the PR is a bit misleading. It reads like we are changing the process.

good point actually

DEVELOPMENT.md
seant-aws: this probably should be, noting for now

kenyrish-amazon (approved):

PR #381, opened 2026-04-09
feat: output filenames customizability with tokens

jairaws:

The previous submitter automatically prepended \<Layer\> and \<Camera\> tokens to the output file prefix based on the number of renderable layers/cameras. This was opaque and surprising — users couldn't control the output directory structure or
filename pattern directly.

Users need full control over the imageFilePrefix pattern, including the ability to use tokens like \<Scene\>, \<RenderLayer\>, and \<Camera\> that get resolved at render time.

- Added a new filename_utils.py module with a SUPPORTED_TOKENS dict (single source of truth) and a resolve_tokens() function that performs case-insensitive token replacement at render time.
- Replaced the auto-prepend logic in renderers.py with a WYSIWYG approach: the pattern is passed through exactly as the user types it.
- Added an "Output File Prefix" group box to the Scene Settings tab with a pattern input field and a live preview that resolves tokens using the current scene context.
- The pattern is stored as a sticky setting (output_file_prefix_pattern in RenderSubmitterUISettings) and applied uniformly to all render layers at submission time.
- All five render handlers (default_maya_handler, arnold_handler, redshift_handler, renderman_handler, vray_handler) now call _resolve_output_file_prefix() instead of passing raw tokens to Maya.
- The <Camera> token resolves to the camera shape node name (e.g., topCamShape1) to match Maya's native <Camera> token behavior.
- Changed from per-layer OutputFilePrefix parameters to a single global OutputFilePrefix parameter in the job template.
- Added OCIOConfigFile parameter to the job template init-data.
- Updated are_images_similar in integ test helper to search recursively for output images, since Maya renders into subdirectories based on layer/camera.

- Users now have explicit control over output filename patterns via the submitter UI.
- The submitter no longer silently modifies the output prefix — what you type is what gets submitted.
- Jobs submitted with older versions of the submitter that used per-layer OutputFilePrefix parameters will not work with this version of the adaptor. However, the adaptor still resolves the same tokens Maya supports, so the rendered output is functionally equivalent.

- Unit tests for resolve_tokens() in test_filename_utils.py
- Unit tests for _resolve_output_file_prefix() in test_maya_handler_base.py, including camera shape node resolution
- Integration tests for token resolution (test_token_resolution_adaptor) and minimal scene rendering (test_minimal_scene_adaptor)
- Submitter integration tests in test_maya_submitters.py
- Job bundle output tests for all scene types

#### Integ test result

PASS
hatch run integ:test_submitters
hatch run integ:test_adaptors_maya (test_token_resolution_adaptor, test_minimal_scene_adaptor).

#### Installer test result

N/A — no changes to installer/.

Timestamp: 2026-04-08T10:51:21.883958-07:00
Running job bundle output test: /Users/ruizjair/workplace/jairaws/deadline-cloud-for-maya/job_bundle_output_tests/layers

layers
Test succeeded

Timestamp: 2026-04-08T10:51:23.717737-07:00
Running job bundle output test: /Users/ruizjair/workplace/jairaws/deadline-cloud-for-maya/job_bundle_output_tests/referenced_layers

referenced_layers
Test succeeded

Timestamp: 2026-04-08T10:51:25.360205-07:00
Running job bundle output test: /Users/ruizjair/workplace/jairaws/deadline-cloud-for-maya/job_bundle_output_tests/renderman
Timestamp: 2026-04-08T10:51:25.360297-07:00
Running job bundle output test: /Users/ruizjair/workplace/jairaws/deadline-cloud-for-maya/job_bundle_output_tests/cube

cube
Test succeeded

Timestamp: 2026-04-08T10:51:26.907453-07:00
Running job bundle output test: /Users/ruizjair/workplace/jairaws/deadline-cloud-for-maya/job_bundle_output_tests/layers_no_variation

layers_no_variation
Test succeeded

Timestamp: 2026-04-08T10:51:32.362913-07:00

- Yes — added docs/design/output-filename-with-tokens.md design document.
- Docstrings added/updated for all new and modified functions.
- Schema files were not modified.

- [ ] Yes
- [x] No

No. The adaptor consumes output_file_prefix as a plain string in init-data — it doesn't know or care about the template parameter names. Old job bundles with per-layer parameters still resolve to valid output_file_prefix strings that the new
adaptor handles correctly. Similarly, new job bundles with the single OutputFilePrefix parameter produce tokens that the old adaptor passes through to Maya, which resolves them natively.

### Default

Screenshot 2026-04-08 at 7 58 36 PM

### Custom Tokens

Screenshot 2026-04-08 at 8 01 52 PM

Screenshot 2026-04-08 at 7 58 46 PM

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/docs/design/output-filename-with-tokens.md
+++ b/docs/design/output-filename-with-tokens.md
@@ -1,0 +1,242 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+# Output File Prefix Token System — Design Document
+
+## Overview
+
+This feature adds output file prefix token resolution to the Deadline Cloud for Maya submitter and adaptor. Users type the full `imageFilePrefix` pattern directly, with optional tokens that get resolved by the adaptor at render time.
+
+**WYSIWYG principle:** The pattern is passed through exactly as the user types it. No automatic prepending of tokens, no automatic cleanup. What you see is what you get.
+
+**Supported tokens:**
+
+- `<Scene>` (alias: `%s`) — Scene file name without extension (e.g., "myScene")
+- `<RenderLayer>` (aliases: `<Layer>`, `%l`) — Render layer display name (e.g., "masterLayer")
+- `<Camera>` (alias: `%c`) — Camera name (e.g., "persp", "renderCam")
+
+Token definitions are maintained in a single source of truth: `SUPPORTED_TOKENS` dict in `filename_utils.py`. The UI tooltip is generated from this dict via `get_tokens_tooltip()`, so adding a new token only requires updating one place.
+
+Everything else in the pattern is literal text. Path separators (`/`) create subdirectories.
+
+**All tokens are resolved by the adaptor at render time.** The submitter passes the raw pattern through to init-data. The adaptor already has all the data it needs: scene file path from init-data, render layer from init-data, and camera from run-data or init-data.
+
+**Unknown tokens pass through unmodified.** Renderer-specific tokens like `<RenderPass>` are left for Maya or the renderer to handle.
+
+**Token matching is case-insensitive.** Both `<Scene>` and `<scene>` resolve to the scene name.
+
+**Example patterns and results:**
+
+| Pattern | Scene=myScene, Layer=fg, Camera=renderCam |
+|---------|-------------------------------------------|
+| `<Scene>/<RenderLayer>/<RenderLayer>` | `myScene/fg/fg` |
+| `<Scene>/<Camera>/<RenderLayer>` | `myScene/renderCam/fg` |
+| `<Camera>/<Scene>` | `renderCam/myScene` |
+| `<Scene>` | `myScene` |
+| `myRender` | `myRender` |
+
+**Default pattern initialization:**
+
+1. Read `defaultRenderGlobals.imageFilePrefix` from the Maya scene
+2. If empty, default to `<Scene>`
+3. Sticky settings always take precedence — if the user previously saved a pattern, that loads instead
+
+---
+
+## 1. Data Structures
+
+### 1.1 New field on `RenderSubmitterUISettings`
+
+**File:** `src/deadline/maya_submitter/data_classes.py`
+
+```python
+@dataclass
+class RenderSubmitterUISettings:
+    ...existing fields...
+
+    # Output file prefix pattern with optional tokens (<Scene>, <RenderLayer>, <Camera>).
+    # Resolved by the adaptor at render time. Empty string means use scene default.
+    output_file_prefix_pattern: str = field(default="", metadata={"sticky": True})
+```
+
+### 1.2 Shared utility functions
+
+**File:** `src/deadline/maya_adaptor/MayaClient/filename_utils.py`
+
+This file contains:
+
+- `SUPPORTED_TOKENS` — single source of truth dict mapping canonical tokens to their aliases
+- `get_tokens_tooltip()` — builds the UI tooltip string from `SUPPORTED_TOKENS`
+- `resolve_tokens()` — case-insensitive token replacement with alias support
+
+```python
+SUPPORTED_TOKENS: dict[str, list[str]] = {
+    "<Scene>": ["%s"],
+    "<RenderLayer>": ["<Layer>", "%l"],
+    "<Camera>": ["%c"],
+}
+
+def resolve_tokens(
+    pattern: str,
+    scene_name: str = "",
+    render_layer: str = "",
+    camera: str = "",
+) -> str:
+    """
+    Replace supported tokens in a Maya imageFilePrefix pattern with actual values.
+    Case-insensitive. Aliases included. Path structure (/) is preserved.
+    Unknown tokens pass through unmodified.
+    """
+    ...
+```
+
+---
+
+## 2. UX Changes (Submitter Dialog)
+
+### 2.1 Output File Prefix pattern field
+
+**File:** `src/deadline/maya_submitter/ui/components/scene_settings_tab.py`
+
+A new "Output File Prefix" group box is added to the scene settings tab, containing the pattern input and a live preview:
+
+```
+Row 0: Project Path       [/path/to/project              ] [...]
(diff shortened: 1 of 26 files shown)


job_bundle_output_tests/cube/expected_job_bundle/template.yaml
```diff
@@ -66,6 +66,14 @@ parameterDefinitions:
   allowedValues:
   - 'true'
   - 'false'
+- name: OCIOConfigFile
```
jairaws, line 69: Test job bundles were failing due to OCIO. I believe this is related to a previous commit.

docs/design/output-filename-with-tokens.md
```diff
@@ -0,0 +1,242 @@
...
+
+This file contains:
+
+- `SUPPORTED_TOKENS` — single source of truth dict mapping canonical tokens to their aliases
+- `get_tokens_tooltip()` — builds the UI tooltip string from `SUPPORTED_TOKENS`
+- `resolve_tokens()` — case-insensitive token replacement with alias support
+
+```python
+SUPPORTED_TOKENS: dict[str, list[str]] = {
+    "<Scene>": ["%s"],
```
loachri, line 73: question: Where does %s come from? Was this something we've had internally?

loachri (approved):
LGTM as a newer reviewer. This passes my skim/smell test, though I'm curious what devs more familiar with this codebase thinks

docs/design/output-filename-with-tokens.md
```diff
@@ -0,0 +1,242 @@
...
+| `src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py` | `_resolve_output_file_prefix()` method; updated `start_render()` |
+| `src/deadline/maya_adaptor/MayaClient/render_handlers/arnold_handler.py` | Uses `_resolve_output_file_prefix()` |
+| `src/deadline/maya_adaptor/MayaClient/render_handlers/vray_handler.py` | Uses `_resolve_output_file_prefix()` |
+| `src/deadline/maya_adaptor/MayaClient/render_handlers/redshift_handler.py` | Uses `_resolve_output_file_prefix()` |
+| `src/deadline/maya_adaptor/MayaClient/render_handlers/renderman_handler.py` | Uses `_resolve_output_file_prefix()` |
+| `src/deadline/maya_submitter/data_classes.py` | Add `output_file_prefix_pattern` sticky field |
+| `src/deadline/maya_submitter/ui/components/scene_settings_tab.py` | "Output File Prefix" group box with pattern + live preview |
+| `src/deadline/maya_submitter/maya_render_submitter.py` | Default pattern from scene; raw pattern passed to job bundle |
+| `src/deadline/maya_submitter/renderers.py` | Removed auto-prepend logic |
+| `test/unit/.../test_filename_utils.py` | Unit tests for `resolve_tokens()` |
```
seant-aws, line 241: these are good, one callout though that caused me a headache, we should double check our separate internal tests, since they are actually separate form the test suite in this repo. with 3dsMax, due to the file output changes, the output filenames actually changed when render tests were ran, leading to test fails; may be worth checking for Maya.

src/deadline/maya_adaptor/MayaClient/filename_utils.py
```diff
@@ -0,0 +1,62 @@
...
+Shared output filename utilities for Deadline Cloud Maya integration.
+Provides token-based output filename prefix resolution.
+
+Supported tokens are resolved by the adaptor at render time.
+Unknown tokens (e.g., renderer-specific ones like <RenderPass>) pass through
+unmodified for Maya or the renderer to handle.
+"""
+import re
+
+SUPPORTED_TOKENS: dict[str, list[str]] = {
```
seant-aws, line 12: This is something new for me too, so might be a callout for other DCCs as well, autodesk just released their mcp server so I used it in part for reviewing as well: 

```https://developer.api.autodesk.com/knowledge/public/v1/mcp```

I added this mcp server, and its got a lot more context for autodesk products now.

Anyway, I notice you said stuff like <RenderPass> are render specific, so we're allowing to pass through, which gets properly handled already.

Just double clicking, I found some more tokens, would these also fall under render specific, so we can allow these to pass through ?
https://help.autodesk.com/view/MAYAUL/2027/ENU/?guid=GUID-F6C6AE73-F484-4534-B992-2E599D4AE91E


Token | What it does
-- | --
Scene> | Scene file name
RenderLayer> | Render layer name (e.g., layer1). With Render Setup, adds layer name to image filename instead of creating subfolders
Camera> | Renderable camera name (e.g., camera1). If rendering fields, appends _odd/_even
RenderPassFileGroup> | Pass group name (e.g., IndirectIllum)
RenderPass> | Render pass node name (e.g., diffuseNoShadow)
RenderPassType> | Abbreviated pass type, <6 chars (e.g., REFR for refraction)
Extension> | File format extension — useful for creating directory structures by format
Version> | Version label (number, date, time, or custom)

docs/design/output-filename-with-tokens.md
seant-aws: it seems like t he old code auto prepended <Camera>/ and <RenderLayer>/ when multiple cameras or layers were detected? but now it might stick to a WYSISWG kinda deal for output filenames, which may lead to test failures

jairaws:
Closing this PR. The rendering tokens are pass to Maya and resolved internally. We do not need to resolve them in the adaptor.

PR #390, opened 2026-04-14
fix: detect OCIO config from environment variable and remap path on worker before scene open

### LINKED ISSUE #379: Bug: OCIO environment variable not detected by submitter - worker renders with wrong/default color config
### Describe the bug

The submitter does not detect or forward the OCIO environment variable. The standard `OCIO` env var is the primary mechanism used by OpenColorIO-aware renderers (V-Ray, Arnold, RenderMan, etc.) to locate the OCIO config file. Pipeline tools such as [AYON](https://help.ayon.app/articles/3005275-core-addon-settings#kerrbx5bbds) and other studio-level tooling commonly set this variable globally.

Currently, `Scene.ocio_config_file()` in `scene.py` only queries `maya.cmds.colorManagementPrefs(query=True, configFilePath=True)`. It does not check `os.environ.get("OCIO")`. When the OCIO config is set via the environment variable rather than Maya's internal preferences, the submitter returns `None` and the `OCIOConfigFile` job parameter is never populated.

On the worker, the `OCIO` env var either isn't set or still points to the submitter machine's path, which doesn't exist on the worker. The renderer then fails to load the OCIO config and falls back to a raw/default config.

Worker error:
`2026/03/30 16:56:17+02:00 ADAPTOR_OUTPUT: STDERR: V-Ray error: Error could not read 'Z:\OCIO\Maya2022-default\config.ocio' OCIO profile.
2026/03/30 16:56:17+02:00 ADAPTOR_OUTPUT: STDERR: V-Ray warning: No valid OCIO config could be loaded - using a default fallback (raw) config instead`

### Expected Behaviour

The submitter should also check the `OCIO` environment variable when detecting the active OCIO config. If set, the path should be resolved, included as the `OCIOConfigFile` parameter, and path-mapped on the worker so the renderer can find it. Alternatively, the adaptor should set the `OCIO` environment variable on the worker with the mapped path before the scene is opened.

### Current Behaviour

1. Pipeline tooling sets OCIO=Z:\OCIO\Maya2022-default\config.ocio as an environment variable
2. V-Ray uses this env var to find the OCIO config on the submitter machine
3. `Scene.ocio_config_file()` only checks `colorManagementPrefs`, doesn't see a custom config, returns `None`
4. `OCIOConfigFile` parameter is never added to the job
5. On the worker, V-Ray reads the stale/unmapped `OCIO` env var (or it's absent), fails to find the config file
6. V-Ray falls back to raw config — render has incorrect color management
7. No warning is shown at submission time

### Reproduction Steps

1. Set the OCIO environment variable to point to an OCIO config file (e.g. `OCIO=Z:\OCIO\Maya2022-default\config.ocio`) - this can be done via pipeline tools like AYON, studio environment setup, or manually.
2. Launch Maya 2024 with V-Ray - V-Ray will use the `OCIO` env var for color management
3. Open a scene and submit via the Deadline Cloud submitter
4. Check the submitted job bundle — OCIOConfigFile will be empty
5. On the worker, V-Ray fails to read the OCIO config at the original path and falls back to raw

### Environment

Operating system: Windows (submitter side)
Version of Maya: 2024
Renderer: V-Ray
Version of this package: 0.15.14 (post PR #361 and #375)
OCIO config: Set via `OCIO` environment variable by pipeline tooling

oehmends:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-maya/issues/379

When the `OCIO` environment variable is set by pipeline tools (e.g. [AYON](https://help.ayon.app/articles/3005275-core-addon-settings#kerrbx5bbds), ShotGrid, or studio-level tooling) to configure color management globally, the submitter and adaptor did not detect or handle it. This caused renders on Deadline Cloud workers to use the wrong color config:

1. `Scene.ocio_config_file()` only checked `maya.cmds.colorManagementPrefs(query=True, configFilePath=True)` — it did not check `os.environ.get("OCIO")`
2. When the OCIO config was set via the environment variable, the submitter returned `None` and the `OCIOConfigFile` job parameter was never populated
3. On the worker, the `OCIO` env var was either absent or still pointed to the submitter machine's Windows path (e.g. `Z:\OCIO\Maya2022-default\config.ocio`)
4. The renderer (V-Ray, Arnold, etc.) failed to load the OCIO config and fell back to a raw/default config, producing incorrect color management in the rendered output

The `OCIO` environment variable is the standard mechanism used by all OpenColorIO-aware renderers to locate the OCIO config file, so this is a common production scenario.

**Submitter side** (`scene.py`):
- `Scene.ocio_config_file()` now falls back to `os.environ.get("OCIO")` when Maya's `colorManagementPrefs` returns a `<MAYA_RESOURCES>` default path or a non-string value (no custom config set). This ensures the `OCIOConfigFile` parameter is populated and the config file is included in the job's asset references for upload.

**Adaptor side** (`adaptor.py`, `default_maya_handler.py`):
- Moved `ocio_config_file` from `_MAYA_INIT_KEYS` (processed after scene open) to a new `_PRE_SCENE_OPTIONAL_KEYS` list (processed before scene open). This is necessary because renderers like V-Ray initialize their color pipeline at scene-open time by reading the `OCIO` env var — if it's not set yet, they fall back to raw.
- `set_ocio_config_file()` now:
  1. Sets `os.environ["OCIO"]` so renderers that read the env var directly pick up the correct config at scene open
  2. Sets `colorManagementPrefs` so Maya's internal color management is also configured
  3. For `.ma` (ASCII) scene files: rewrites the `.cfp` attribute (`defaultColorMgtGlobals.configFilePath`) in the scene file before opening, replacing the unmapped Windows path with the mapped worker path. This prevents Maya from logging errors about the unmapped path during `maya.cmds.file(open=True)`.
  4. Re-applies `colorManagementPrefs` after scene open as a safety net, since `maya.cmds.file(open=True)` can override the global prefs with the path embedded in the scene file.

For `.mb` (binary) scene files, the scene file patching is skipped (binary format), but the `OCIO` env var and `colorManagementPrefs` are still set, so the render result is correct — only a cosmetic Maya warning about the unmapped path remains in the log.

Jobs submitted from environments where the `OCIO` env var is set by pipeline tooling will now render with the correct color management on Deadline Cloud workers. Previously these jobs silently rendered with wrong/default color config.

This is not a breaking change. The submitter still checks `colorManagementPrefs` first and only falls back to the env var when no usable custom config is found there. Jobs that don't use the `OCIO` env var are unaffected.

- Reproduced the issue: set `OCIO` env var on submitter machine, submitted a V-Ray job, confirmed the worker fell back to raw config
- After the fix: confirmed `OCIOConfigFile` is populated in the job bundle, the OCIO config file is uploaded, and the worker renders with correct color management
- Tested with both `.ma` and `.mb` scene files
- Tested with V-Ray renderer (reads `OCIO` env var directly)

#### Integ test result

N/A — no changes to integration test infrastructure.

The OCIO config detection change does not affect job bundle output structure — it only changes whether the `OCIOConfigFile` parameter value is populated (which depends on the runtime environment having the `OCIO` env var set). The existing job bundle output tests do not set the `OCIO` env var, so they are not affected.

- Updated docstrings on `Scene.ocio_config_file()` and `set_ocio_config_file()` to reflect the new behavior
- No README or schema changes needed

- [ ] Yes
- [x] No

No. The submitter still checks `colorManagementPrefs` first and only falls back to the `OCIO` env var. The adaptor change (pre-scene action ordering) is internal and does not modify the init-data or run-data schema. Jobs submitted with older versions of the submitter will continue to work.

----

### DIFF
--- a/src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
+++ b/src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
@@ -56,6 +56,7 @@ class DefaultMayaHandler:
         self.camera_name = None
         self.output_file_prefix = None
         self.render_kwargs = {}
+        self._pending_ocio_path = None
 
     def get_camera_to_render(self, data: dict) -> list[str]:
         # The ls function returns all of the camera shapes, but the cameras themselves are represented by
@@ -279,6 +280,15 @@ class DefaultMayaHandler:
             raise FileNotFoundError(f"The scene file '{file_path}' does not exist")
         maya.cmds.file(file_path, open=True, force=True, ignoreVersion=ignore_version_flag)
 
+        # Re-apply OCIO config after scene open — maya.cmds.file(open=True)
+        # overrides colorManagementPrefs with the path embedded in the scene file
+        if self._pending_ocio_path:
+            print(
+                f"Re-applying OCIO config after scene open: '{self._pending_ocio_path}'", flush=True
+            )
+            maya.cmds.colorManagementPrefs(e=True, configFilePath=self._pending_ocio_path)
+            self._pending_ocio_path = None
+
         pre_render_mel = maya.cmds.getAttr("defaultRenderGlobals.preMel")
         if pre_render_mel:
             try:
@@ -290,8 +300,10 @@ class DefaultMayaHandler:
         """
         Sets the OCIO config file path for color management.
 
-        This should be called after the scene file is opened to override
-        the scene's embedded OCIO config path with the remapped path.
+        This is called before the scene file is opened so that both
+        Maya's color management and renderers like V-Ray (which read
+        the OCIO environment variable) pick up the correct config
+        at scene-open time.
 
         Args:
             data (dict): The data given from the Adaptor. Keys expected: ['ocio_config_file']
@@ -308,6 +320,13 @@ class DefaultMayaHandler:
             print(f"WARNING: OCIO config file not found: '{ocio_path}'", flush=True)
             return
 
-        # Set the OCIO config file path in Maya's color management preferences
-        print(f"Setting OCIO config: '{ocio_path}'", flush=True)
+        # Set the OCIO environment variable so renderers pick it up at scene open
+        os.environ["OCIO"] = ocio_path
+
+        # Set Maya's color management prefs before scene open to prevent
+        # Maya from trying to load the unmapped path embedded in the scene file
         maya.cmds.colorManagementPrefs(e=True, configFilePath=ocio_path)
+
+        # Store path so we can re-apply after scene open (scene open may override)
+        self._pending_ocio_path = ocio_path
+        print(f"Setting OCIO config: '{ocio_path}'", flush=True)
--- a/job_bundle_output_tests/cube/expected_job_bundle/template.yaml
+++ b/job_bundle_output_tests/cube/expected_job_bundle/template.yaml
@@ -97,10 +97,6 @@ parameterDefinitions:
     groupLabel: Maya Settings
   minValue: 1
   description: The image height of the output.
-jobEnvironments:
-- name: Set OCIO Config Path
-  variables:
-    OCIO: '{{Param.OCIOConfigFile}}'
 steps:
 - name: masterLayer
   parameterSpace:
--- a/job_bundle_output_tests/cube/expected_job_bundle/template.yaml
+++ b/job_bundle_output_tests/cube/expected_job_bundle/template.yaml
@@ -97,6 +97,10 @@ parameterDefinitions:
     groupLabel: Maya Settings
   minValue: 1
   description: The image height of the output.
+jobEnvironments:
+- name: Set OCIO Config Path
+  variables:
+    OCIO: '{{Param.OCIOConfigFile}}'
 steps:
 - name: masterLayer
   parameterSpace:
--- a/job_bundle_output_tests/layers/expected_job_bundle/template.yaml
+++ b/job_bundle_output_tests/layers/expected_job_bundle/template.yaml
@@ -232,6 +232,10 @@ parameterDefinitions:
   allowedValues:
   - 'true'
   - 'false'
+jobEnvironments:
+- name: Set OCIO Config Path
+  variables:
+    OCIO: '{{Param.OCIOConfigFile}}'
 steps:
 - name: layerDefaultFrames
   parameterSpace:
--- a/job_bundle_output_tests/layers/expected_job_bundle/template.yaml
+++ b/job_bundle_output_tests/layers/expected_job_bundle/template.yaml
@@ -232,10 +232,6 @@ parameterDefinitions:
   allowedValues:
   - 'true'
   - 'false'
-jobEnvironments:
-- name: Set OCIO Config Path
-  variables:
-    OCIO: '{{Param.OCIOConfigFile}}'
 steps:
 - name: layerDefaultFrames
   parameterSpace:
--- a/job_bundle_output_tests/layers_no_variation/expected_job_bundle/template.yaml
+++ b/job_bundle_output_tests/layers_no_variation/expected_job_bundle/template.yaml
@@ -108,6 +108,10 @@ parameterDefinitions:
   allowedValues:
   - 'true'
   - 'false'
+jobEnvironments:
+- name: Set OCIO Config Path
+  variables:
+    OCIO: '{{Param.OCIOConfigFile}}'
 steps:
 - name: layerTheFIrst
   parameterSpace:
--- a/job_bundle_output_tests/layers_no_variation/expected_job_bundle/template.yaml
+++ b/job_bundle_output_tests/layers_no_variation/expected_job_bundle/template.yaml
@@ -108,10 +108,6 @@ parameterDefinitions:
   allowedValues:
   - 'true'
   - 'false'
-jobEnvironments:
-- name: Set OCIO Config Path
-  variables:
-    OCIO: '{{Param.OCIOConfigFile}}'
 steps:
 - name: layerTheFIrst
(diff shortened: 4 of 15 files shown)


oehmends:
### Optional follow-up: `.ma` scene file patching to suppress Maya OCIO log errors

During development of this fix, I also implemented a pre-scene-open patch that rewrites the `.cfp` attribute (`defaultColorMgtGlobals.configFilePath`) inside `.ma` (ASCII) scene files before `maya.cmds.file(open=True)`. This replaces the unmapped Windows path (e.g. `Z:\OCIO\Maya2022-default\config.ocio`) with the mapped worker path directly in the scene file text.

I intentionally left this out of the PR because it is **cosmetic only** — it suppresses Maya's error log messages about the unmapped OCIO path during scene open, but does not affect render correctness. The `OCIO` env var (set by this PR) is what renderers actually read, and the post-scene-open `colorManagementPrefs` re-application ensures Maya's internal state is correct.

**What it would look like:**
- Before `maya.cmds.file(open=True)`, read the `.ma` file and regex-replace `setAttr ".cfp" -type "string" "<old_path>";` with the mapped path
- Only applies to `.ma` files (`.mb` binary files cannot be patched this way)
- Falls back gracefully if the file can't be read/written

**Trade-offs:**
- ✅ Eliminates Maya OCIO error messages from worker logs (cleaner logs)
- ⚠️ Modifies the scene file on disk (worker-side copy only)
- ⚠️ Only works for `.ma` files, not `.mb`
- ⚠️ Adds complexity for a cosmetic improvement

src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
```diff
@@ -308,6 +318,13 @@ def set_ocio_config_file(self, data: dict) -> None:
             print(f"WARNING: OCIO config file not found: '{ocio_path}'", flush=True)
             return
 
-        # Set the OCIO config file path in Maya's color management preferences
-        print(f"Setting OCIO config: '{ocio_path}'", flush=True)
+        # Set the OCIO environment variable so renderers pick it up at scene open
+        os.environ["OCIO"] = ocio_path
```
kenyrish-amazon, line 324: I'd suggest also adding a jobEnvironments entry in [`_get_job_template()`](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/maya_render_submitter.py#L270) to set the OCIO env   
  var in the template, matching how the Nuke submitter does it via [`_add_ocio_path_to_job_template()`](https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/mainline/src/deadline/nuke_submitter/deadline_submitter_for_nuke.py#L141):                                                                                              
      
                                                                                                                                                                                                                    
```
  if ocio_config:                                                                                                                                                                                                                  
      if "jobEnvironments" not in job_template:                                                                                                                                                                                    
          job_template["jobEnvironments"] = []                                                                                                                                                                                     
      job_template["jobEnvironments"].insert(0, {                                                                                                                                                                                  
          "name": "Set OCIO Config Path",                                                                                                                                                                                          
          "variables": {"OCIO": "{{Param.OCIOConfigFile}}"},                                                                                                                                                                       
      })                                                                                                                                                                                                                           
                 
```                                                                                                                                                                                                                 
  This makes the template self-contained — if someone exports the bundle and resubmits it standalone, the OCIO env var still gets set. Since OCIOConfigFile is a PATH type with dataFlow: IN, Deadline Cloud's path mapping handles remapping automatically. The adaptor-side os.environ + colorManagementPrefs handling should still stay for the Maya-specific scene-open behavior.

kenyrish-amazon (commented):
Thanks for the PR! This looks good, and would fix the OCIO config issue. Just left a comment about adding the variable to jobEnvironments as well.

src/deadline/maya_adaptor/MayaClient/render_handlers/default_maya_handler.py
oehmends: Thanks for the suggestion! I initially added the `jobEnvironments` entry, but after looking at how the Nuke submitter handles this, I think the Maya case is different enough that it's not needed here.

**How Nuke does it:** The Nuke adaptor (`nuke_handler.py`) has no OCIO handling at all — no `os.environ["OCIO"]`, no color management prefs. The `jobEnvironments` entry in the Nuke submitter is the *only* mechanism that sets the `OCIO` env var on the worker. Without it, Nuke workers would have no OCIO config.

**How Maya does it:** The Maya adaptor already has an explicit `set_ocio_config_file()` action that:
1. Sets `os.environ["OCIO"]` before scene open
2. Sets `colorManagementPrefs` before scene open
3. Re-applies `colorManagementPrefs` after scene open (since `maya.cmds.file(open=True)` overrides it)

This action is triggered via the `OCIOConfigFile` parameter flowing through `init_data` → adaptor → handler. The adaptor handles the full lifecycle including path mapping via `DirectoryMapping.convert()`.

Adding a `jobEnvironments` entry on top of this would set `OCIO` at job environment setup time, and then the adaptor would overwrite it again during `set_ocio_config_file()` — making it redundant.

I've reverted the `jobEnvironments` commit. If you think there's still value in having it as a belt-and-suspenders approach (e.g. for standalone bundle resubmission without the adaptor), happy to add it back.

kenyrish-amazon: That makes sense! Having it in both places is not strictly needed here, thanks for explaining the reasoning. This looks good to me, but I see one of the checks failed due to formatting issues. I can approve after that is fixed!

kenyrish-amazon (dismissed):

kenyrish-amazon (approved):
LGTM!

jairaws (approved):

PR #394, opened 2026-04-21
Add job bundle to run integ tests in Deadline Cloud

jairaws:
Running the integration requires setting up Maya and its plugins in the user environment. The environment needs to have licenses configured for all the plugins (when it applies). Thus, running the full set of integration tests as part of the development cycle can be complicated.

Create a job bundle to submit a job that runs the integ tests in an real fleet. The job bundle is configured to pull the plugins from the default conda channel.

Easier to run the integ tests.

Submitted the job and verified it passed.

Yes, added README file.

- [ ] Yes
- [x] No

No
----

### DIFF
--- a/job_bundle_integ_tests/README.md
+++ b/job_bundle_integ_tests/README.md
@@ -1,0 +1,39 @@
+# Maya Integration Tests on Deadline Cloud
+
+Runs the deadline-cloud-for-maya integration tests on a Deadline Cloud Service Managed Fleet with all supported renderer plugins (Arnold, V-Ray, Redshift).
+
+## Prerequisites
+
+- A Deadline Cloud farm with a queue that has the default Conda queue environment.
+- The `deadline` CLI installed and configured (`deadline config`).
+
+## Usage
+
+```bash
+cd /path/to/deadline-cloud-for-maya
+deadline bundle submit job_bundle_integ_tests \
+  --name "maya-integ-tests" \
+  --parameter "RepoDir=." \
+  --parameter "CondaPackages=maya=2026 maya-mtoa=2026.5.5 maya-vray=2026.7 maya-redshift=2026.2" \
+  --max-retries-per-task 0
+```
+
+## Steps
+
+| Step | Description |
+|------|-------------|
+| `submitter` | Runs submitter tests (`test_maya_submitters.py`). Validates that the Maya submitter generates correct job bundles. |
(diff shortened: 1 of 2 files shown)


job_bundle_integ_tests/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Prerequisites
+
+- A Deadline Cloud farm with a queue that has the default Conda queue environment.
+- The `deadline` CLI installed and configured (`deadline config`).
+
+## Usage
+
+```bash
+deadline bundle submit /path/to/this/job-bundle \
```
mwiebe: If you start from the deadline-cloud-for-maya dir, you could give a copy-pastable command. E.g.

```
cd /path/to/deadline-cloud-for-maya
deadline bundle submit job_bundle_integ_tests \
    -p RepoDir=. \
    ...
```

job_bundle_integ_tests/template.yaml
```diff
@@ -0,0 +1,108 @@
...
+      runnable: true
+      data: |
+        #!/bin/bash
+        set -euo pipefail
+        cd '{{Param.RepoDir}}'
+        mayapy -m pytest test/integ/test_maya_submitters.py -vvv --numprocesses=1 --no-cov -m submitter
+    actions:
+      onRun:
+        command: '{{Task.File.Run}}'
+        timeout: 600
```
mwiebe, line 93: Would this benefit from some host requirements? Or does this work on all OS?

job_bundle_integ_tests/README.md
jairaws: Good catch. Updated the readme.

job_bundle_integ_tests/template.yaml
jairaws: yeah, makes sense... the job is meant for linux... added host requirements...

loachri (approved):

job_bundle_integ_tests/README.md
```diff
@@ -0,0 +1,39 @@
+# Maya Integration Tests on Deadline Cloud
```
kenyrish-amazon, line 1: Do we need to update this section as well for running integ tests? https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/DEVELOPMENT.md#integration-tests

kenyrish-amazon (approved):

Issue #384, opened 2026-04-09
Bug: Layers and Cameras dropdowns not populating


yuanmich2:
### Describe the bug

<img width="1097" height="971" alt="Image" src="https://github.com/user-attachments/assets/3dcd0b57-7cc4-443a-a80f-00c0aa34da26" />

The submitter only shows "Current Layer" and "All Renderable Layers" in the Layers dropdown instead of populating with the actual layer names from the scene. The same issue is present for the Cameras dropdown.

### Expected Behaviour

The dropdowns should populate with the actual layer and camera names from the scene.

### Current Behaviour

The dropdowns only show All or Current layer/camera options.

### Reproduction Steps

Create a scene with multiple cameras and layers, then open the submitter and try to select one of the layers/cameras.

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.)
2. Version of Maya
3. Version of this package
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer?

Please share other details about your environment that you think might be relevant to reproducing the bug.

yuanmich2:
Additional screenshot to show what the Camera dropdown looks like. My original post was a little bit inaccurate. I assumed it'd be the same as the Layers dropdown, but the Camera dropdown doesn't show a "Current camera" option. I can only select the "All Renderable Cameras" option.

<img width="1267" height="589" alt="Image" src="https://github.com/user-attachments/assets/16887e0d-af86-4594-8634-75bb701cb889" />

kenyrish-amazon:
Fixed this in the PR linked, thanks for finding this bug!

PR #402, opened 2026-05-05
fix: populate cameras dropdown when submitter dialog opens (#384)

kenyrish-amazon:
## What was the problem

The Cameras dropdown in the Maya submitter UI only showed "All Renderable Cameras" instead of listing individual scene cameras. This prevented artists from selecting a specific camera for their render job.

**Root cause:** Commit 321adbd refactored `show_maya_render_submitter` by moving `on_create_job_bundle_callback` from a nested closure to a module-level function. During this refactoring, the camera population code was moved INTO `on_create_job_bundle_callback` (which only runs at submit time) but was NOT kept in `show_maya_render_submitter` (which runs at dialog open time). As a result, the camera lists retained their default values of `[ALL_CAMERAS]` when the dialog opened.

GitHub issue: https://github.com/aws-deadline/deadline-cloud-for-maya/issues/384

## What was the solution

Added the camera population code back into `show_maya_render_submitter()`, after `_set_render_layer_data()` returns and before the dialog is created. This mirrors the same logic already in `on_create_job_bundle_callback`.

Also fixed `test_frame_override_has_text_validation` which was broken since its introduction (commit 5467cbf) due to:
1. Using `del sys.modules[...]` on a module that was never imported (KeyError)
2. Not mocking transitive dependencies (maya.*, qtpy.QtCore, qtpy.QtGui, deadline.client.ui)

## How was this change tested

- Unit tests: 39 submitter tests pass, 79 adaptor tests pass
- Linting: ruff check + black --check pass
- The previously-failing test_frame_override_has_text_validation now passes

 Manual testing in Maya 2026:
  
  1. Created scene with 3 cameras (camera1, camera2, camera3) + persp, all marked renderable
  2. Created 2 render layers (layer_A, layer_B) with per-layer camera overrides via Render Setup (camera3 disabled on layer_A, camera2 disabled on layer_B)
  3. Opened submitter → Camera dropdown shows all 4 cameras ✅ (previously only showed "All Renderable Cameras")
  4. Switched Layers to "All Renderable Layers" → Camera dropdown shows only camera1 and persp (intersection) ✅
  5. Switched Layers to "Current Layer" → Camera dropdown shows camera1, camera2, persp (current layer's renderable cameras) ✅

image

image

### DIFF
--- a/src/deadline/maya_submitter/maya_render_submitter.py
+++ b/src/deadline/maya_submitter/maya_render_submitter.py
@@ -45,6 +45,30 @@ import time
 logger = getLogger(__name__)
 
 
+def _populate_selectable_cameras(
+    render_settings: "RenderSubmitterUISettings",
+    render_layers: list["RenderLayerData"],
+) -> None:
+    """Populate the selectable camera lists on render_settings for the UI dropdowns.
+
+    - current_layer_selectable_cameras: all renderable cameras in the current layer
+    - all_layer_selectable_cameras: only cameras common to ALL render layers (intersection)
+    """
+    current_layer_selectable_cameras: list[str] = get_renderable_camera_names()
+    render_settings.current_layer_selectable_cameras = [ALL_CAMERAS] + sorted(
+        current_layer_selectable_cameras
+    )
+
+    all_layer_selectable_cameras_set: set[str] = set(render_layers[0].renderable_camera_names)
+    for layer in render_layers:
+        all_layer_selectable_cameras_set = all_layer_selectable_cameras_set.intersection(
+            layer.renderable_camera_names
+        )
+    render_settings.all_layer_selectable_cameras = [ALL_CAMERAS] + sorted(
+        all_layer_selectable_cameras_set
+    )
+
+
 @dataclass
 class RenderLayerData:
     name: str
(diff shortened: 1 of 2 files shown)


test/unit/deadline_submitter_for_maya/scripts/test_submitter.py
```diff
@@ -39,5 +78,151 @@ def __init__(self, parent):
...
+            ["cam_front", "cam_side", "cam_top"],  # layer A
+            ["cam_front", "cam_side"],  # layer B (missing cam_top)
+        ]
+        all_layer_set = set(layer_cameras[0])
+        for cameras in layer_cameras:
+            all_layer_set = all_layer_set.intersection(cameras)
+        render_settings.all_layer_selectable_cameras = [ALL_CAMERAS] + sorted(all_layer_set)
+
+        # Current layer mode: all cameras from current layer
+        assert render_settings.current_layer_selectable_cameras == [
```
jairaws, line 161: We are mocking the values of `current_layer_selectable_cameras` and `all_layer_selectable_cameras` directly in the test. Thus, asserting them doesn't seem to be validating the underlying functionality.

We need to verify the values of these variables when `show_maya_render_submitter` is called.

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -689,6 +689,20 @@ def update_progress(message):
...
     all_renderers: set[str] = {layer_data.renderer_name for layer_data in render_layers}
 
+    # Populate the selectable cameras for the UI dropdowns
+    current_layer_selectable_cameras: list[str] = get_renderable_camera_names()
+    render_settings.current_layer_selectable_cameras = [ALL_CAMERAS] + sorted(
+        current_layer_selectable_cameras
+    )
+    all_layer_selectable_cameras_set: set[str] = set(render_layers[0].renderable_camera_names)
+    for layer in render_layers:
+        all_layer_selectable_cameras_set = all_layer_selectable_cameras_set.intersection(
```
jairaws: I believe that we are trying to calculate all unique camera names across all layers, correct? If so, I don't see how getting the intersection of cameras across all layers accomplishes this. Maybe the intent here is different than what I thought? Could we add a comment to clarify what camera names we intent to collect here?

test/unit/deadline_submitter_for_maya/scripts/test_submitter.py
kenyrish-amazon: Fixed — extracted _populate_selectable_cameras() helper and rewrote tests to call it directly with mocked inputs. Tests now exercise the real code path.

src/deadline/maya_submitter/maya_render_submitter.py
kenyrish-amazon: It's intersection by design — this logic already existed in on_create_job_bundle_callback before the regression. When "All Renderable Layers" is selected, the chosen camera must be valid on every layer, so we only show cameras common to all. I added a docstring clarifying this.

jairaws (approved):

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -45,6 +45,30 @@
...
+
+    - current_layer_selectable_cameras: all renderable cameras in the current layer
+    - all_layer_selectable_cameras: only cameras common to ALL render layers (intersection)
+    """
+    current_layer_selectable_cameras: list[str] = get_renderable_camera_names()
+    render_settings.current_layer_selectable_cameras = [ALL_CAMERAS] + sorted(
+        current_layer_selectable_cameras
+    )
+
+    all_layer_selectable_cameras_set: set[str] = set(render_layers[0].renderable_camera_names)
```
ttblanchard: Do we guarantee to always have one render layer? If not we risk index out of bound error here.

sakshie95 (approved):

PR #405, opened 2026-05-08
feat: Extract submitter logic with SubmissionContext pattern

leon-li-inspire:
## Change
Refactor the submission pipeline to separate expensive Maya scene queries from data transformation. Introduces `SubmissionContext` to cache scene data and avoid redundant `_set_render_layer_data()` calls when multiple public functions are used together.

This makes parts of the submitter logic externally callable, allowing integrations (e.g. AYON) to get submission data as dictionaries before they are dumped to job bundle YAML files.

### Public API for external integrations

```python
from deadline.maya_submitter.maya_render_submitter import (
    create_submission_context,
    get_job_template_for_submission,
    get_parameter_values_for_submission,
    get_asset_references_for_submission,
    get_queue_parameters,
    get_job_bundle_for_submission,
)
from deadline.maya_submitter.data_classes import RenderSubmitterUISettings
from deadline.client.job_bundle.submission import AssetReferences


settings = RenderSubmitterUISettings()
queue_parameters: list[dict[str, Any]] = get_queue_parameters()
asset_references = AssetReferences(
    input_filenames=set(settings.input_filenames),
    input_directories=set(settings.input_directories),
    output_directories=set(settings.output_directories),
)

# Recommended: single entry point
job_bundle = get_job_bundle_for_submission()

# Or use individual functions with shared context to avoid redundant scene queries:
context = create_submission_context(settings)
job_template = get_job_template_for_submission(settings, context=context)
parameter_values = get_parameter_values_for_submission(settings, queue_parameters, context=context)
asset_refs_dict = get_asset_references_for_submission(asset_references)
```

#### Note about `get_queue_parameters()`
This function sets Open Job Description queue parameters manually (not from the UI).

```python
# With initial values
queue_params = get_queue_parameters(
    initial_values={"CondaPackages": "maya=2026.* maya-mtoa"}
)

# Explicitly (won't get farm_id and queue_id from settings)
queue_params = get_queue_parameters(
    farm_id="farm-xxxxx",
    queue_id="queue-xxxx",
    initial_values={"CondaPackages": "maya=2026.* maya-mtoa"}
)
```

## Test plan
- [x] Run `hatch run test` to verify unit tests pass
- [x] Run `hatch run install --maya-version 2026` and verify plugin loads in Maya
- [x] Test `get_job_bundle_for_submission()` from Maya Script Editor
- [x] Test individual functions with shared `SubmissionContext`
- [x] Verify existing submitter UI workflow still works end-to-end

### DIFF
--- a/src/deadline/maya_submitter/maya_render_submitter.py
+++ b/src/deadline/maya_submitter/maya_render_submitter.py
@@ -476,9 +476,7 @@ def _get_parameter_values(
             )
 
     parameter_values.extend(
-        {"name": param["name"], "value": param.get("value", param.get("default", ""))}
-        for param in queue_parameters
-        if "value" in param or "default" in param
+        {"name": param["name"], "value": param["value"]} for param in queue_parameters
     )
 
     return parameter_values
@@ -593,10 +591,18 @@ def create_submission_context() -> SubmissionContext:
     )
 
 
+@dataclass
+class PreparedRenderLayers:
+    """Result of filtering and annotating render layers for submission."""
+
+    layers: list[RenderLayerData]
+    renderers: set[str]
+
+
 def _prepare_render_layers_for_submission(
     settings: RenderSubmitterUISettings,
     context: SubmissionContext,
-) -> tuple[list[RenderLayerData], set[str]]:
+) -> PreparedRenderLayers:
     """Filter and annotate render layers based on settings.
 
     This is a pure transformation — no Maya scene queries. It works on
@@ -608,7 +614,7 @@ def _prepare_render_layers_for_submission(
         context: Pre-computed submission context.
 
     Returns:
-        A tuple of (submit_render_layers, renderers).
+        PreparedRenderLayers with the filtered layers and renderer set.
     """
     render_layers = deepcopy(context.render_layers)
 
@@ -630,6 +636,8 @@ def _prepare_render_layers_for_submission(
         layer.frame_range != first_frame_range for layer in submit_render_layers
     )
 
+    # If there are multiple frame ranges and we're not overriding the range,
+    # then we create per-layer Frames parameters.
     if per_layer_frames_parameters:
         for layer_data in submit_render_layers:
             layer_data.frames_parameter_name = f"{layer_data.display_name}Frames"
@@ -657,7 +665,7 @@ def _prepare_render_layers_for_submission(
 
     renderers: set[str] = {layer_data.renderer_name for layer_data in submit_render_layers}
 
-    return submit_render_layers, renderers
+    return PreparedRenderLayers(layers=submit_render_layers, renderers=renderers)
 
 
 def get_default_job_template() -> dict[str, Any]:
@@ -687,13 +695,13 @@ def get_job_template_for_submission(
         context = create_submission_context()
 
     default_job_template = get_default_job_template()
-    submit_render_layers, renderers = _prepare_render_layers_for_submission(settings, context)
+    prepared = _prepare_render_layers_for_submission(settings, context)
 
     job_template = _get_job_template(
         default_job_template=default_job_template,
         settings=settings,
-        renderers=renderers,
-        render_layers=submit_render_layers,
+        renderers=prepared.renderers,
+        render_layers=prepared.layers,
         all_layer_selectable_cameras=context.all_layer_selectable_cameras,
         current_layer_selectable_cameras=context.current_layer_selectable_cameras,
     )
@@ -729,9 +737,9 @@ def get_parameter_values_for_submission(
     if queue_parameters is None:
         queue_parameters = []
 
-    submit_render_layers, renderers = _prepare_render_layers_for_submission(settings, context)
+    prepared = _prepare_render_layers_for_submission(settings, context)
 
-    return _get_parameter_values(settings, renderers, submit_render_layers, queue_parameters)
+    return _get_parameter_values(settings, prepared.renderers, prepared.layers, queue_parameters)
 
 
 def get_asset_references_for_submission(
@@ -749,9 +757,9 @@ def get_asset_references_for_submission(
 
 
 def get_queue_parameters(
-    farm_id: Optional[str] = None,
-    queue_id: Optional[str] = None,
-    initial_values: Optional[dict[str, Any]] = None,
+    farm_id_override: Optional[str] = None,
+    queue_id_override: Optional[str] = None,
+    initial_values_override: Optional[dict[str, Any]] = None,
 ) -> list[dict[str, Any]]:
     """Get queue parameters from Deadline Cloud for external API usage.
 
@@ -761,9 +769,9 @@ def get_queue_parameters(
     going through the UI.
 
     Args:
-        farm_id: The farm ID. If not provided, uses the default from settings.
-        queue_id: The queue ID. If not provided, uses the default from settings.
-        initial_values: Optional dict of {parameter_name: value} to override
+        farm_id_override: The farm ID. If not provided, uses the default from settings.
+        queue_id_override: The queue ID. If not provided, uses the default from settings.
+        initial_values_override: Optional dict of {parameter_name: value} to override
             default parameter values.
 
     Returns:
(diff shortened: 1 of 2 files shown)


src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -618,24 +633,229 @@ def on_create_job_bundle_callback(
...
+        context = create_submission_context()
+    if queue_parameters is None:
+        queue_parameters = []
+
+    submit_render_layers, renderers = _prepare_render_layers_for_submission(settings, context)
+
+    return _get_parameter_values(settings, renderers, submit_render_layers, queue_parameters)
+
+
+def get_asset_references_for_submission(
```
crowecawcaw, line 747: One idea: if we want to make this a standard pattern, we could make an abstract class in the deadline library and then have each submitter extend it. It could be a soft way of pushing submitters to using common interfaces and patterns. We'd need to make sure it wasn't too rigid because there will surely be exceptions and odd cases.

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -435,18 +439,22 @@ def _get_parameter_values(
...
             conda_param["value"] = " ".join(
-                pkg for pkg in conda_param["value"].split() if not pkg.startswith("maya-openjd")
+                pkg for pkg in current_value.split() if not pkg.startswith("maya-openjd")
             )
 
     parameter_values.extend(
-        {"name": param["name"], "value": param["value"]} for param in queue_parameters
+        {"name": param["name"], "value": param.get("value", param.get("default", ""))}
+        for param in queue_parameters
+        if "value" in param or "default" in param
```
jairaws, line 481: Q: Why are we adding this filtering logic on the parameters? From the existing implementation, it seems like "value" will always exist. Otherwise this logic would throw an exception.

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -589,8 +606,6 @@ def on_create_job_bundle_callback(
         layer.frame_range != first_frame_range for layer in submit_render_layers
     )
 
-    # If there are multiple frame ranges and we're not overriding the range,
```
jairaws, line 604: Q: Why did we decide to remove this comment? Is it no longer true? Should we update it instead of removing it?

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -519,58 +527,67 @@ def _set_render_layer_data() -> list[RenderLayerData]:
...
+        all_layer_selectable_cameras=all_layer_selectable_cameras,
+    )
 
-    job_bundle_path = Path(job_bundle_dir)
 
-    # If we're only submitting the current layer, filter our list of layers by that
+def _prepare_render_layers_for_submission(
+    settings: RenderSubmitterUISettings,
+    context: SubmissionContext,
+) -> tuple[list[RenderLayerData], set[str]]:
```
jairaws: nit: Could we create a data class for the output as well?

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -618,24 +633,229 @@ def on_create_job_bundle_callback(
...
+        asset_references: The asset references object.
+
+    Returns:
+        The asset references dictionary ready for serialization.
+    """
+    return asset_references.to_dict()
+
+
+def get_queue_parameters(
+    farm_id: Optional[str] = None,
```
jairaws: nit: We could add the suffix "override" to the param names to make clear their intent.

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -519,58 +527,67 @@ def _set_render_layer_data() -> list[RenderLayerData]:
...
+    """
+    render_layers = _set_render_layer_data()
+    current_layer_selectable_cameras = get_renderable_camera_names()
+
     all_layer_selectable_cameras_set: set[str] = set(render_layers[0].renderable_camera_names)
     for layer in render_layers:
         all_layer_selectable_cameras_set = all_layer_selectable_cameras_set.intersection(
             layer.renderable_camera_names
         )
-    all_layer_selectable_cameras: list[str] = list(sorted(all_layer_selectable_cameras_set))
```
kenyrish-amazon: [This PR](https://github.com/aws-deadline/deadline-cloud-for-maya/pull/402) moved getting the camera/layers to its own function, we should rebase with this.

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -618,24 +633,229 @@ def on_create_job_bundle_callback(
...
-    # If "HostRequirements" is provided, inject it into each of the "Step"
     if host_requirements:
-        # for each step in the template, append the same host requirements.
         for step in job_template["steps"]:
             step["hostRequirements"] = host_requirements
 
+    return job_template
+
+
+def get_parameter_values_for_submission(
```
kenyrish-amazon, line 718: Should we add unit tests for each of these functions?

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -519,58 +527,67 @@ def _set_render_layer_data() -> list[RenderLayerData]:
...
+        settings: The render submitter UI settings.
+        context: Pre-computed submission context.
+
+    Returns:
+        A tuple of (submit_render_layers, renderers).
+    """
+    render_layers = deepcopy(context.render_layers)
+
     if settings.render_layer_selection == LayerSelection.CURRENT:
         current_render_layer_name = get_current_render_layer_name()
```
kenyrish-amazon, line 624: Should current_render_layer_name be cached as well?

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: It is because a difference in the two code paths:

  1. UI path (on_create_job_bundle_callback): The queue_parameters come from the Qt submitter dialog as list[JobParameter]. The UI widget always ensures each parameter dict has a "value" key set by the user.
  2. External/API path (get_queue_parameters()): Calls get_queue_parameter_definitions() which returns parameter definitions from the Deadline Cloud API. These dicts have "name", "type", and "default" — but no "value" key unless the
  caller explicitly sets one via initial_values_override. Only parameters that the user overrides get "value" added.

So when AYON calls get_queue_parameters() without initial_values_override, the returned params only have "default", not "value". The filter if "value" in param or "default" in param with the fallback param.get("value", param.get("default", "")) handles both cases.

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: Sure, will add in the next rev.

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: I checked this method, seems it has been only called in this place, seems fine to not cache for now. We can absolutely to cache this later when necessary.

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: Will add it back in the next rev.

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: Yea, this is our ultimate goal, currently we are still collecting requirements from AYON side for https://github.com/aws-deadline/deadline-cloud/issues/1078

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: Added the unit tests

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: Added in the next rev.

jairaws (approved):

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: I have rebased

loachri (approved):

PR #410, opened 2026-05-25
test: add OCIO config file submission test coverage

loachri:

In `0.15.13`, customers without a custom OCIO config in their Maya scene experienced "Job bundle validation failed" submission errors. Their `OCIOConfigFile` job parameter relied on its empty-string default, which was rejected by the version of `openjd-model-for-python` shipped at the time (HIDDEN PATH parameters were not allowed to have empty-string defaults).

The bug was fixed in `0.15.14` by upgrading the dependency, but the developer test suite never exercised the customer-default code path — every existing scene fixture had Maya's default `<MAYA_RESOURCES>/...` color management value, which `Scene.ocio_config_file()` correctly resolves to `None`, so `OCIOConfigFile` is not added to `parameter_values` and the parameter falls back to its empty-string default. The bug was therefore invisible to existing tests.

This PR adds the missing regression coverage so a future regression of the same shape is caught by CI.

**Unit tests** in `test/unit/deadline_submitter_for_maya/scripts/test_submission_api.py`, new `TestOCIOInJobBundle` class:

| Test | Covers |
|---|---|
| `test_scenario_1_no_ocio_omits_value_but_keeps_definition` | Scene without OCIO config — `OCIOConfigFile` is omitted from `parameter_values`, parameter definition keeps `default: ''` |
| `test_scenario_2_ocio_present_adds_value_to_parameter_values` | Scene with OCIO config — value populated from `Scene.ocio_config_file()` |
| `test_scenario_3_hidden_path_with_empty_default_validates` | HIDDEN PATH with empty default validates via `deadline-cloud`'s `validate_job_parameter` (the function that surfaces "Job bundle validation failed" errors) |
| `test_default_template_yaml_ocio_param_validates` | Loads the bundled `default_maya_job_template.yaml` and asserts every parameter validates — guards against future template edits regressing the OCIO contract |
| `test_full_bundle_with_no_ocio_passes_parameter_validation` | Full bundle assembly + parameter validation end-to-end — the closest unit-level reproduction of the original failure |

**Integration test** in `test/integ/test_maya_submitters.py` — new `test_scene_submitter_with_ocio_config` method, with fixture under `test/integ/test_scripts/ocio_test/`:

* `scene/test.ma` — Arnold-based scene (copy of the existing `mtoa_test`)
* `scene/config.ocio` — minimal valid OCIO v2 config (single linear colorspace, no LUTs)
* `expected_job_bundle/template.yaml` — identical to `mtoa_test`'s template (the OCIO state lives in `parameter_values`, not the template)

The test programmatically applies a custom OCIO config via `cmds.colorManagementPrefs(e=True, configFilePath=...)` post-scene-open (simulating a customer with a studio-wide OCIO config in Maya's preferences) and asserts `OCIOConfigFile` is present in the resulting `parameter_values` with the expected path.

The four pre-existing parametrized integration tests (`Minimal Maya Test`, `Redshift Test`, `Arnold Test`, `VRay Test`) implicitly cover the customer-default Scenario 1 case — their scenes use the `<MAYA_RESOURCES>/...` default which `Scene.ocio_config_file()` correctly returns `None` for. The new dedicated method completes the matrix with Scenario 2.

Tests-only change. No `src/` code is modified, so no runtime behaviour changes.

**Side observation** surfaced while writing the integration test: once `colorManagementPrefs(configFilePath=...)` is set, Maya's `filePathEditor` reports the OCIO config file as a scene-referenced file, so `AssetIntrospector.parse_scene_assets()` adds it to `inputs.filenames` in addition to the dedicated `OCIOConfigFile` `dataFlow:IN` PATH parameter. Job Attachments deduplicates by content hash so this results in a single S3 upload, but the OCIO config is referenced twice in the bundle representation. This is pre-existing behaviour, not introduced by this PR — the integration test now codifies it.

```
$ hatch run test
============================== 183 passed in 5.50s ==============================

$ hatch run lint
ruff check . — All checks passed!
black --check --diff . — All done! ✨ 🍰 ✨ — 81 files would be left unchanged.
mypy src test maya_submitter_plugin/plug-ins — Success: no issues found in 72 source files
```

#### Integ test result

Integration tests pass on Linux and Windows (CodeBuild reserved-capacity fleets, manual `workflow_dispatch`):

```
========================== short test summary info ===========================
Captured:
- test_minimal_scene_adaptor passed
- test_redshift_scene_adaptor passed
- test_mtoa_scene_adaptor passed
- test_vray_scene_adaptor passed
- test_scene_submitter[Minimal Maya Test] passed
- test_scene_submitter[Redshift Test] passed
- test_scene_submitter[Arnold Test] passed
- test_scene_submitter[VRay Test] passed
- test_scene_submitter_with_ocio_config passed   <-- new
======================= 9 passed, 4 warnings in ~2:14 ========================
```

#### Installer test result

Not applicable — `installer/` was not modified.

Not applicable. This PR places the integration fixture under `test/integ/test_scripts/` (the path used by the automated CI integration tests on CodeBuild) rather than `job_bundle_output_tests/` (the in-Maya `TEST` shelf workflow). The same OCIO scenario could be added to `job_bundle_output_tests/` in a follow-up if the dev workflow needs it; it is not required to land the regression coverage that motivated this PR.

No documentation changes are needed — this PR adds tests; public APIs and adaptor schemas are not modified. The `CHANGELOG.md` does not need a hand-edited entry (semantic-release classifies `test:` commits as no-version-bump).

- [ ] Yes
- [x] No

   - [ ] Yes
   - [ ] No

No. Tests-only.

----

### DIFF
--- a/test/integ/test_maya_submitters.py
+++ b/test/integ/test_maya_submitters.py
@@ -251,3 +251,171 @@ class TestSubmitters:
         are_asset_references_similar(
             job_history_dir, expected_asset_references, job_configuration.expected_scene_file_paths
         )
+
+    def test_scene_submitter_with_ocio_config(
+        self,
+        initialize_maya,
+        script_location: Path,
+        tmp_path: Path,
+    ) -> None:
+        """OCIO config file regression coverage.
+
+        Companion to the 0.15.13 OCIO submission failure that was fixed
+        in 0.15.14 (the customer-default code path was uncovered by
+        existing tests).
+
+        The four parametrized tests above implicitly cover the
+        no-custom-OCIO case (scene's ``cfp`` is the default
+        ``<MAYA_RESOURCES>/...`` path which ``Scene.ocio_config_file()``
+        correctly returns ``None`` for, so ``OCIOConfigFile`` is omitted
+        from parameter_values and falls back to its empty-string default
+        in the template). That's the customer state that broke when
+        ``openjd-model-for-python < 0.9.1`` rejected empty defaults on
+        HIDDEN PATH parameters.
+
+        This test covers the custom-OCIO case — scene **with** a custom
+        OCIO config. We programmatically apply the config via
+        ``cmds.colorManagementPrefs(e=True, configFilePath=...)``
+        post-scene-open (simulating a customer with a studio-wide OCIO
+        config set in Maya's preferences). The submitter must then pick
+        it up via ``Scene.ocio_config_file()`` and emit it as the
+        ``OCIOConfigFile`` parameter value in the bundle.
+        """
+        show_maya_render_submitter = initialize_maya[0]
+        on_create_job_bundle_callback = initialize_maya[1]
+
+        asset_folder = "ocio_test"
+        file_prefix = "ocio_test"
+        frame_list = "1"
+
+        job_history_dir = tmp_path / "jobhistory"
+        output_path = tmp_path / "output"
+        project_path = script_location / asset_folder / "scene"
+        scene_location = project_path / "test.ma"
+        ocio_config_path = project_path / "config.ocio"
+
+        # Sanity check the fixture exists before doing any Maya work.
+        assert ocio_config_path.is_file(), f"OCIO config fixture missing at {ocio_config_path}"
+
+        # Clean up sticky setting
+        self._cleanup_sticky_settings(scene_location, script_location)
+
+        os.makedirs(job_history_dir, exist_ok=True)
+        os.makedirs(output_path, exist_ok=True)
+
+        cmds.workspace(project_path, openWorkspace=True)
+        cmds.workspace(fileRule=["images", output_path])
+
+        cmds.file(scene_location, open=True, force=True)
+
+        cmds.setAttr("defaultResolution.width", 960)
+        cmds.setAttr("defaultResolution.height", 540)
+        cmds.setAttr("defaultRenderGlobals.imageFilePrefix", file_prefix, type="string")
+        cmds.optionVar(iv=("renderSetup_includeAllLights", 0))
+
+        # The crux of this test: apply a custom OCIO config to Maya's
+        # color management preferences. Scene.ocio_config_file() reads
+        # this back via colorManagementPrefs(query=True, configFilePath=True).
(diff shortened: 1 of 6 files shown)


test/integ/test_maya_submitters.py
```diff
@@ -251,3 +251,171 @@ def test_scene_submitter(
...
+        script_location: Path,
+        tmp_path: Path,
+    ) -> None:
+        """OCIO config file regression coverage.
+
+        Companion to the 0.15.13 OCIO submission failure that was fixed
+        in 0.15.14 (the customer-default code path was uncovered by
+        existing tests).
+
+        The four parametrized tests above implicitly cover the
```
TanoVip123, line 267: Q: Out of curiosity, if the 4 parameterized test cases above already covered the case of no-custom-OCIO, why did we not catch that before it broke customer?

test/integ/test_maya_submitters.py
```diff
@@ -251,3 +251,171 @@ def test_scene_submitter(
...
+
+        os.makedirs(job_history_dir, exist_ok=True)
+        os.makedirs(output_path, exist_ok=True)
+
+        cmds.workspace(project_path, openWorkspace=True)
+        cmds.workspace(fileRule=["images", output_path])
+
+        cmds.file(scene_location, open=True, force=True)
+
+        cmds.setAttr("defaultResolution.width", 960)
```
TanoVip123, line 311: nit: I notice we use these hardcoded number "960", "540", "20" across all integ submitter tests. Would it be a good idea to parameterize them?

loachri: It's a fair callout, though I think it's okay to not parametrize them since this is mostly for a test, and it's fairly clear what these numbers correspond to (i.e. we know that 960 corresponds to the output width

test/integ/test_maya_submitters.py
loachri: That's a good question! Maybe the comment could be better here, but I think the above tests didn't fully test OCIO handling, only partially. The tests above don't invoke the validator like these tests here.

TanoVip123 (approved):

ttblanchard (approved):

PR #435, opened 2026-07-02
feat: implement MayaSubmitter on the unified BaseSubmitter

leon-li-inspire:
## Summary
Implements `MayaSubmitter` against the unified `BaseSubmitter` ABC (aws-deadline/deadline-cloud#1245), so Maya can be driven through the shared submission contract in addition to the existing native GUI submitter. The submitter is stateful and the module layout is split into a headless engine and a GUI layer.

## Changes
- **`maya_submitter/submitter.py`** (renamed from `submitter_api.py`) — headless engine, no Qt:
  - `MayaSubmitter(BaseSubmitter)`: `get_settings` / `get_job_template` / `get_parameter_values` / `get_asset_references` (+ inherited `get_submission_context`). The two large builders (`_get_job_template` / `_get_parameter_values`) are `@staticmethod`s on the class.
  - Stateful, query-once: the expensive scene scan (render-layer switching, camera enumeration) is cached lazily as a `MayaSceneContext` on the instance and reused across `get_job_template` and `get_parameter_values`, so a submission queries the scene once. `reset_scene_context()` clears it; the constructor accepts an injected context.
  - `create_scene_context()` builds the context; `_get_render_layer_data()` gathers per-layer data; `_prepare_render_layers_for_submission()` does the CURRENT-layer filtering + per-layer parameter naming (multi-layer fix — see below).
- **`MayaSubmitterSettings(BaseSubmitterSettings)`** gains `include_adaptor_wheels` + `timeouts`; `render_layer_selection` is typed as `LayerSelection` end-to-end. Subclassing the base is what lets host-agnostic consumers (e.g. the AYON bridge) treat Maya's settings polymorphically.
- **`maya_render_submitter.py`** keeps GUI-only code plus a `_ui_settings_to_submitter_settings` adapter (`name` → `job_name`). It builds one `MayaSubmitter` and shares its cached context across camera-list population and both builder calls.
- **`get_asset_references` returns the typed `AssetReferences`** (not a dict), matching the finalized `BaseSubmitter` contract; callers serialize with `.to_dict()` at the job-bundle boundary.

## Review feedback addressed
- Moved `_get_job_template` / `_get_parameter_values` into `MayaSubmitter` as staticmethods; moved `_prepare_render_layers_for_submission` onto the class too. `create_scene_context` / `_get_render_layer_data` stay module-level (the GUI uses them directly, and `create_scene_context` is the factory whose result the GUI injects into `MayaSubmitter`).
- Renamed `create_submission_context` → `create_scene_context` and `_set_render_layer_data` → `_get_render_layer_data`.
- Removed `get_maya_queue_parameters`; callers use `deadline.client.api.get_queue_parameters`.
- `get_settings` now logs a warning (instead of a silent `pass`) when `defaultResolution` can't be read.
- Multi-layer correctness (earlier bot findings): renderers are derived **after** the CURRENT filter; `_prepare_render_layers_for_submission` assigns per-layer `Frames` / `OutputFilePrefix` / `ImageWidth` / `ImageHeight` parameter names so differing layers no longer collapse onto layer 0's values; a not-renderable CURRENT layer raises a user-facing `DeadlineOperationError` rather than an opaque `IndexError`.

## Requires
`deadline >= 0.60.2` (pinned in `pyproject.toml`) — the release that ships the unified `BaseSubmitter` / `BaseSubmitterSettings` / `get_queue_parameters` and the typed `get_asset_references` (aws-deadline/deadline-cloud#1245).

## Compatibility with pre-GUI hooks (#1255)
Rebased/merged on current `mainline`, which added the pre-GUI hooks feature to `show_maya_render_submitter`. Both integrate cleanly: the merged GUI entrypoint runs pre-GUI hooks and then drives the stateful `MayaSubmitter`.

## Testing
- Verified against the **released** `deadline` 0.60.2: `mypy`, `ruff`, and `black` all clean; the full `deadline_submitter_for_maya` unit suite passes (incl. the pre-GUI-hook tests) under `pytest-xdist`.
- CI's Python matrix (3.9–3.12 × ubuntu/macos), DCO, and Semantic PR are green.
- In-Maya farm submission verified end-to-end against this stateful refactor: GUI publish -> MayaSubmitter -> CreateJob rendered successfully on a live farm.


src/deadline/maya_submitter/submitter_api.py
```diff
@@ -0,0 +1,163 @@
...
+        settings.output_path = Scene.output_path() or ""
+        settings.frame_list = str(Animation.frame_list())
+        settings.renderer = Scene.renderer() or ""
+
+        try:
+            width = maya.cmds.getAttr("defaultResolution.width")
+            height = maya.cmds.getAttr("defaultResolution.height")
+            settings.image_width = int(width)
+            settings.image_height = int(height)
+        except Exception:
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in 5e2adfd — the bare `except` now carries an explanatory comment plus `# noqa: BLE001`; it intentionally falls back to the default resolution when the scene resolution can't be read.

src/deadline/maya_submitter/submitter_api.py
```diff
@@ -0,0 +1,163 @@
...
+            all_layer_selectable_cameras,
+            current_layer_selectable_cameras,
+        ) = _collect_job_template_inputs()
+
+        if (
+            isinstance(settings, MayaSubmitterSettings)
+            and settings.render_layer_selection == "CURRENT"
+        ):
+            current_name = get_current_render_layer_name()
+            render_layers = [layer for layer in render_layers if layer.name == current_name]
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in 5e2adfd — `_filter_layers_for_selection` now raises a user-facing `DeadlineOperationError` ("… is not set as renderable") instead of producing an empty layer list, and both `get_job_template` and `get_parameter_values` route through it. Covered by `test_filter_layers_current_not_renderable_raises`.

src/deadline/maya_submitter/submitter_api.py
```diff
@@ -0,0 +1,163 @@
...
+        return job_template
+
+    def get_parameter_values(
+        self,
+        settings: SubmitterSettings,
+        queue_parameters: list[dict[str, Any]],
+    ) -> list[dict[str, Any]]:
+        maya_settings = self._to_native_settings(settings)
+        render_layers = _set_render_layer_data()
+        renderers = {layer.renderer_name for layer in render_layers}
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in 5e2adfd — `renderers` is now derived from the filtered `render_layers` (after the CURRENT narrowing), so a non-submitted layer's renderer no longer leaks into the parameter values. Covered by `test_get_parameter_values_renderers_scoped_to_current_layer`.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,826 @@
...
+            )
+            print(f"_set_render_layer_data done processing layer {time.time()}")
+
+    # Sort the layers by name
+    render_layers.sort(key=lambda layer: layer.display_name)
+
+    return render_layers
+
+
+def create_submission_context() -> MayaSceneContext:
```
leon-li-inspire: Should this be renamed to `create_scene_context` ?

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,826 @@
...
+    """Maya-specific submission settings.
+
+    Extends the DCC-agnostic :class:`BaseSubmitterSettings` with the fields the
+    Maya job-template/parameter builders consume directly, so the builders can
+    read a single settings object without a native-settings translation step.
+    """
+
+    # LayerSelection (not a str): the builders compare this against the enum, so
+    # the canonical type flows end-to-end. External callers that never touch it
+    # (e.g. AYON) simply inherit the ALL default.
```
leon-li-inspire: Don't mention AYON in the comment

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,826 @@
...
+    renderers: set[str]
+
+
+def get_default_job_template() -> dict[str, Any]:
+    """Returns the default job template for Maya render submissions."""
+    with open(Path(__file__).parent / "default_maya_job_template.yaml") as fh:
+        return yaml.safe_load(fh)
+
+
+def _set_render_layer_data() -> list[RenderLayerData]:
```
leon-li-inspire: Why this method's name is not `_get_render_layer_data` ?

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,826 @@
...
+    parameter_values.extend(
+        {"name": param["name"], "value": param.get("value", param.get("default", ""))}
+        for param in queue_parameters
+        if "value" in param or "default" in param
+    )
+
+    return parameter_values
+
+
+def get_maya_queue_parameters(
```
leon-li-inspire: This method seems redundant to me, the external library could use the one in `deadline.client.api`. Remove this method.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,826 @@
...
+        settings.frame_list = str(Animation.frame_list())
+        settings.renderer = Scene.renderer() or ""
+
+        try:
+            width = maya.cmds.getAttr("defaultResolution.width")
+            height = maya.cmds.getAttr("defaultResolution.height")
+            settings.image_width = int(width)
+            settings.image_height = int(height)
+        except Exception:  # noqa: BLE001
+            # Resolution is optional metadata: if defaultResolution is missing or
```
leon-li-inspire: Shouldn't we have some logs here ?

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,826 @@
...
+        for layer_data in submit_render_layers:
+            layer_data.image_width_parameter_name = f"{layer_data.display_name}ImageWidth"
+            layer_data.image_height_parameter_name = f"{layer_data.display_name}ImageHeight"
+
+    renderers: set[str] = {layer_data.renderer_name for layer_data in submit_render_layers}
+
+    return PreparedRenderLayers(layers=submit_render_layers, renderers=renderers)
+
+
+def _get_job_template(
```
leon-li-inspire: Should this method be moved into `MayaSubmitter` ?

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,826 @@
...
+        if "jobEnvironments" not in job_template:
+            job_template["jobEnvironments"] = []
+        job_template["jobEnvironments"].append(override_environment["environment"])
+
+    add_timeouts_to_job_template(job_template, settings.timeouts)
+
+    return job_template
+
+
+def _get_parameter_values(
```
leon-li-inspire: Should this method be moved into `MayaSubmitter` ?

test/unit/deadline_submitter_for_maya/plugins/test_deadline_submitter_maya.py
```diff
@@ -10,11 +10,22 @@
...
-import deadline
-from deadline.maya_submitter import maya_render_submitter
-from deadline.maya_submitter import mel_commands
-from deadline.maya_submitter import shelf
-import DeadlineCloudForMaya
+# maya_render_submitter/mel_commands now import the unified submitter engine
+# (deadline.maya_submitter.submitter), which imports BaseSubmitter from
+# deadline-cloud (PR #1245). Skip cleanly until a release carrying it is
+# installed, matching test_submitter.py; auto-activates once the base ships.
+_client_api = pytest.importorskip("deadline.client.api")
```
leon-li-inspire: Don't skip, just leave it as and it's fine if the tests failed.

test/unit/deadline_submitter_for_maya/scripts/test_submission_api.py
```diff
@@ -1,10 +1,28 @@
...
+
 import sys
 from unittest.mock import Mock, patch
 
 import pytest
 
+# The unified BaseSubmitter base class lives in deadline-cloud (PR #1245).
+# ``submitter.py`` imports it at module load, so skip cleanly until a release
+# carrying it is installed (mirrors test_submitter.py).
+_client_api = pytest.importorskip("deadline.client.api")
```
leon-li-inspire: Don't skip, just leave it as and it's fine if the tests failed.

test/unit/deadline_submitter_for_maya/test_mel_commands.py
```diff
@@ -1,9 +1,24 @@
...
-from qtpy.QtWidgets import (  # type: ignore
+
+import pytest
+
+# mel_commands imports maya_render_submitter, which now imports the unified
+# submitter engine (deadline.maya_submitter.submitter). That engine imports
+# BaseSubmitter from deadline-cloud (PR #1245). Skip cleanly until a release
+# carrying it is installed, matching test_submitter.py; auto-activates once the
+# base ships.
+_client_api = pytest.importorskip("deadline.client.api")
```
leon-li-inspire: Don't skip, just leave it as and it's fine if the tests failed.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Done in d1a2744 — renamed `create_submission_context` → `create_scene_context` (+ all call sites and test patch targets).

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Done in d1a2744 — removed the AYON reference from the comment.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Good catch — renamed `_set_render_layer_data` → `_get_render_layer_data` in d1a2744 (it returns data, doesn't set anything). Updated the GUI import and test patch targets too.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Removed in d1a2744 — `get_maya_queue_parameters` is deleted; callers use `deadline.client.api.get_queue_parameters` directly (nothing but its own tests referenced it). Dropped the `TestGetMayaQueueParameters` tests.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Added in d1a2744 — the `except` now emits `logger.warning(..., exc_info=True)` (module-level `getLogger(__name__)`, matching repo convention) before falling back to the default resolution, instead of a silent pass.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Moved in d1a2744 — `_get_job_template` is now a `@staticmethod` on `MayaSubmitter`; `get_job_template` calls `self._get_job_template(...)`.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Moved in d1a2744 — `_get_parameter_values` is now a `@staticmethod` on `MayaSubmitter`; `get_parameter_values` calls `self._get_parameter_values(...)`.

test/unit/deadline_submitter_for_maya/plugins/test_deadline_submitter_maya.py
leon-li-inspire: Done in d1a2744 — removed the skip guard; the module now fails at import until deadline-cloud ships `BaseSubmitter` (PR #1245), and passes once it does.

test/unit/deadline_submitter_for_maya/scripts/test_submission_api.py
leon-li-inspire: Done in d1a2744 — removed the skip guard here as well.

test/unit/deadline_submitter_for_maya/test_mel_commands.py
leon-li-inspire: Done in d1a2744 — removed the skip guard here as well.

leon-li-inspire:
**CI status (intentional):** the `Python` lint/test jobs are red and will stay red until aws-deadline/deadline-cloud#1245 ships `BaseSubmitter`/`BaseSubmitterSettings`. The pinned PyPI `deadline` in CI has no `BaseSubmitter`, so:
- `submitter.py`'s `from deadline.client.api import BaseSubmitter, BaseSubmitterSettings` fails mypy's `attr-defined` check, and
- the submitter test modules fail at import.

Per review, these are deliberately **not** guarded with skips/`type: ignore` — they should fail until the base is available and pass automatically once it is. DCO and Semantic PR are green. No action needed on the red `Python` checks before #1245 lands.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,784 @@
...
+    renderers: set[str]
+
+
+def get_default_job_template() -> dict[str, Any]:
+    """Returns the default job template for Maya render submissions."""
+    with open(Path(__file__).parent / "default_maya_job_template.yaml") as fh:
+        return yaml.safe_load(fh)
+
+
+def _get_render_layer_data() -> list[RenderLayerData]:
```
leon-li-inspire, line 110: Shouldn't the method be moved into the `MayaSubmitter` ?

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,784 @@
...
+            )
+            print(f"_get_render_layer_data done processing layer {time.time()}")
+
+    # Sort the layers by name
+    render_layers.sort(key=lambda layer: layer.display_name)
+
+    return render_layers
+
+
+def create_scene_context() -> MayaSceneContext:
```
leon-li-inspire, line 161: Shouldn't the method be moved into the `MayaSubmitter` ?

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,784 @@
...
+    all_layer_selectable_cameras = [ALL_CAMERAS] + sorted(all_layer_selectable_cameras_set)
+
+    return MayaSceneContext(
+        render_layers=render_layers,
+        current_layer_selectable_cameras=current_layer_selectable_cameras,
+        all_layer_selectable_cameras=all_layer_selectable_cameras,
+    )
+
+
+def _prepare_render_layers_for_submission(
```
leon-li-inspire: Shouldn't the method be moved into the `MayaSubmitter` ?

test/unit/deadline_submitter_for_maya/__init__.py
```diff
@@ -9,11 +9,16 @@
...
+    "maya.app",
+    "maya.app.renderSetup",
+    "maya.app.renderSetup.model",
     "maya.app.renderSetup.model.renderSetupPreferences",
+    "maya.app.general",
     "maya.app.general.fileTexturePathResolver",
     "maya.app.general.mayaMixin",
     "maya.OpenMaya",
     "maya.cmds",
+    "maya.mel",
```
leon-li-inspire, line 21: Why do we need to make the changes in this file ?

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Kept module-level (c34b227). `_get_render_layer_data` is also called by the GUI (`maya_render_submitter.show_maya_render_submitter`) to build its camera/attachment lists without a `MayaSubmitter` instance — it's a free-standing scene query, not a submitter-only method. Making it private would force the GUI to reach through a throwaway instance.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Kept module-level (c34b227). `create_scene_context` is the factory the GUI uses to build a `MayaSceneContext` that it then *injects* into `MayaSubmitter(scene_context=...)`. Keeping it a free function avoids a bootstrapping cycle (needing a submitter to build the context you construct the submitter with) and lets the GUI/tests build a context independently.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Done in c34b227 — moved `_prepare_render_layers_for_submission` onto `MayaSubmitter` as a `@staticmethod` (only the class + its tests use it). Both builder methods now call `self._prepare_render_layers_for_submission(...)`; tests call it via the class.

test/unit/deadline_submitter_for_maya/__init__.py
leon-li-inspire: Once the skip guards were removed, these modules actually import `submitter.py` → `render_layers.py`, which does `import maya.mel` and `import maya.app.renderSetup.model.renderSetupPreferences`. The old mock list only had the leaf `renderSetupPreferences` and was missing `maya.mel` plus the intermediate `maya.app`/`maya.app.renderSetup*`/`maya.app.general` packages, so collection failed with `ModuleNotFoundError: No module named 'maya.mel'`. These additions make the import graph resolve under the mocks.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,789 @@
...
+        settings = MayaSubmitterSettings()
+        settings.job_name = Path(scene_name).name or "Untitled"
+        settings.project_path = Scene.project_path() or ""
+        settings.output_path = Scene.output_path() or ""
+        settings.frame_list = str(Animation.frame_list())
+        settings.renderer = Scene.renderer() or ""
+
+        try:
+            width = maya.cmds.getAttr("defaultResolution.width")
+            height = maya.cmds.getAttr("defaultResolution.height")
```
moonyuet: You dont need to do try and exception, it is always there in Maya.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,789 @@
...
+            settings.image_height = int(height)
+        except Exception:  # noqa: BLE001
+            # Resolution is optional metadata: if defaultResolution is missing or
+            # unreadable (unusual scene state), keep the MayaSubmitterSettings
+            # defaults rather than failing the whole scene scan.
+            logger.warning(
+                "Could not read defaultResolution.width/height from the scene; "
+                "falling back to the default image resolution.",
+                exc_info=True,
+            )
```
moonyuet:
```suggestion
	        settings.image_width = maya.cmds.getAttr("defaultResolution.width")
	        settings.image_height =maya.cmds.getAttr("defaultResolution.height")

```

src/deadline/maya_submitter/submitter.py
moonyuet: Image

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,789 @@
...
+
+        settings.input_filenames = [scene_name] if scene_name else []
+        settings.input_directories = [settings.project_path] if settings.project_path else []
+        settings.output_directories = [settings.output_path] if settings.output_path else []
+
+        return settings
```
moonyuet: I believe we can return to Maya Submitter Settings in this case if there is no try exception.
e.g.
```python
return MayaSubmitterSettings(
    job_name = Path(scene_name).name or "Untitled",
    project_path = Scene.project_path() or "",
    output_path = Scene.output_path() or "",
    
    #....
    # blah blah blah....
)
```

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -74,418 +67,38 @@ def _populate_selectable_cameras(
...
-        {"name": param["name"], "value": param.get("value", param.get("default", ""))}
-        for param in queue_parameters
-        if "value" in param or "default" in param
+    The mirror of the old ``_to_native_settings`` (now inverted): the builders
+    consume MayaSubmitterSettings directly, so the GUI settings are converted at
+    the submission boundary. Only ``name`` -> ``job_name`` needs renaming; every
+    other field maps by name, and ``render_layer_selection`` is already a
+    LayerSelection on both sides.
+    """
+    return MayaSubmitterSettings(
```
moonyuet, line 83: ~This is errored out for now as the relevant dataclass doesn't have that much arguments. You might need to~ ~add the attributes back~
I just find out the related attributes have been implemented in the BaseSubmitterSettings.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,789 @@
...
+        contract; callers serialize it with ``.to_dict()`` at the job-bundle
+        boundary.
+        """
+        scene_file = Scene.name() or ""
+        project_path = Scene.project_path() or ""
+
+        return AssetReferences(
+            input_filenames={scene_file} if scene_file else set(),
+            input_directories={project_path} if project_path else set(),
+            output_directories=set(settings.output_directories),
```
moonyuet: Small question about this: As I saw the submitter settings from the getter of the settings, input filenames, input directories, output directories are with list type values(See: https://github.com/leon-li-inspire/deadline-cloud-for-maya/blob/8df9844d36aa4b5f3391176d329216ecd970adc3/src/deadline/maya_submitter/submitter.py#L253-L255), But now here the AssetReference dataclass treated them as set type. 
I believe there won't be duplicate and we shouldn't allow the duplicate values of those filenames/directories. 
Should we align the values to set types?

src/deadline/maya_submitter/submitter.py
BigRoy: I wonder if someone gets their scene in such a state that `defaultResolution` can't be queried having an actual error might be warranted. But, I'd say at this point this is code style preference. It's not "wrong" and since it does log with `exc_info=True` it should still log what's the cause.

Not a fan of the bare `Exception` myself, but again - it's a preference.

But e.g. it's very hard to get the `defaultResolution` node to not exist I think.
```python
import maya.api.OpenMaya as om

sel = om.MSelectionList()
sel.add("defaultResolution")
node = sel.getDependNode(0)
om.MGlobal.deleteNode(node)
```
Can't delete it either.

src/deadline/maya_submitter/submitter_api.py
```diff
@@ -0,0 +1,188 @@
...
+        (
+            default_job_template,
+            render_layers,
+            all_layer_selectable_cameras,
+            current_layer_selectable_cameras,
+        ) = _collect_job_template_inputs()
+
+        render_layers = self._filter_layers_for_selection(settings, render_layers)
+
+        renderers = {layer.renderer_name for layer in render_layers}
```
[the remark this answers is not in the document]
leon-li-inspire: Addressed — the unified path now routes through `_prepare_render_layers_for_submission`, so per-layer `Frames`/`OutputFilePrefix`/`ImageWidth`/`ImageHeight` params are generated for differing layers. `_collect_job_template_inputs` was removed.

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -679,6 +679,45 @@ def get_default_job_template() -> dict[str, Any]:
...
+    paths stay in sync. Returns raw camera names (no ``ALL_CAMERAS`` sentinel);
+    unlike :func:`create_submission_context` it also tolerates an empty scene.
+
+    Returns:
+        (default_job_template, render_layers, all_layer_selectable_cameras,
+        current_layer_selectable_cameras)
+    """
+    default_job_template = get_default_job_template()
+
+    render_layers = _set_render_layer_data()
```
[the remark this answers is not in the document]
leon-li-inspire: Moot — `_collect_job_template_inputs` (and its empty-scene docstring) was removed in the refactor; the context is built by `create_scene_context`.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Done — `_get_job_template` is now a `@staticmethod` on `MayaSubmitter`.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Done — `_get_parameter_values` is now a `@staticmethod` on `MayaSubmitter`.

test/unit/deadline_submitter_for_maya/plugins/test_deadline_submitter_maya.py
leon-li-inspire: Done — skip guard removed; the module fails (rather than skips) until the base is present, and passes now that the pin requires deadline 0.60.2.

test/unit/deadline_submitter_for_maya/scripts/test_submission_api.py
leon-li-inspire: Done — skip guard removed here as well.

test/unit/deadline_submitter_for_maya/test_mel_commands.py
leon-li-inspire: Done — skip guard removed here as well.

src/deadline/maya_submitter/submitter_api.py
```diff
@@ -0,0 +1,186 @@
...
+@dataclass
+class MayaSubmitterSettings(SubmitterSettings):
+    """Maya-specific submission settings."""
+
+    render_layer_selection: str = "ALL"
+    camera_selection: str = "All Renderable Cameras"
+    renderer: str = ""
+    image_width: int = 0
+    image_height: int = 0
+    output_file_prefix: str = ""
```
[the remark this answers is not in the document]
leon-li-inspire: Addressed in 86609f9 — dropped `renderer`/`image_width`/`image_height`/`output_file_prefix` from `MayaSubmitterSettings` (option A1). They were write-only: `get_settings()` set them but no builder read them. Removing them also removed the dead `defaultResolution` try/except and the unused `maya.cmds` import.

src/deadline/maya_submitter/submitter_api.py
```diff
@@ -0,0 +1,186 @@
...
+        return render_layers
+
+    def get_asset_references(self, settings: SubmitterSettings) -> dict[str, Any]:
+        scene_file = Scene.name() or ""
+        project_path = Scene.project_path() or ""
+
+        asset_refs = AssetReferences(
+            input_filenames={scene_file} if scene_file else set(),
+            input_directories={project_path} if project_path else set(),
+            output_directories=set(settings.output_directories),
```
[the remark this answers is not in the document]
leon-li-inspire: Addressed in 86609f9 — `get_asset_references` now runs `AssetIntrospector.parse_scene_assets()` (the same scan the native GUI submitter uses), classifying detected paths into input files vs directories (dropping non-existent ones), and unions `settings.output_directories` with each render layer's `output_directories` from the cached scene context. So inputs include textures/references/caches and outputs include per-layer locations — matching the native path.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,789 @@
...
+        """Collect asset references (inputs/outputs) from the scene.
+
+        Returns the typed :class:`AssetReferences` per the BaseSubmitter
+        contract; callers serialize it with ``.to_dict()`` at the job-bundle
+        boundary.
+        """
+        scene_file = Scene.name() or ""
+        project_path = Scene.project_path() or ""
+
+        return AssetReferences(
```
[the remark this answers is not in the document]
leon-li-inspire: Addressed in 86609f9 — the headless `get_asset_references` now runs `AssetIntrospector.parse_scene_assets()` (a Qt-free scene util, so the engine stays GUI-free), so textures / referenced scenes / caches are included, not just the scene file. Per-layer output dirs are merged in from the cached scene context. Covered by the rewritten `TestGetAssetReferences`.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Resolved in 86609f9 — the whole `defaultResolution` try/except is gone. Those fields (`image_width`/`image_height`) were write-only (never read by any builder), so they were dropped along with the `getAttr` block. No bare `except` remains.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Resolved in 86609f9 — the `defaultResolution.getAttr` calls were removed entirely (the `image_width`/`image_height` fields they fed were unused), so the suggestion no longer applies. Thanks @BigRoy for confirming the node is effectively undeletable — that reinforced that the try/except was unnecessary.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Done in d354e03 — `get_settings` now returns a single `MayaSubmitterSettings(...)` constructor call instead of instantiate-then-assign, as suggested.

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: Resolved — as you noted, those fields are provided by `BaseSubmitterSettings`, so the adapter constructs `MayaSubmitterSettings(...)` with them directly and no longer errors.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Working as intended — `BaseSubmitterSettings` declares `input_filenames`/`input_directories`/`output_directories` as `list[str]` (the contract), so `get_settings` keeps lists. `AssetReferences` uses `set` in its own dataclass; the list→set conversion happens at the `get_asset_references` boundary (86609f9), which now builds sets directly from the introspector — so there's no duplicate risk. Aligning `get_settings` to `set` would break the base type.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,806 @@
...
+            )
+
+        # If this developer option is enabled, merge the adaptor_override_environment
+        if settings.include_adaptor_wheels:
+            with open(Path(__file__).parent / "adaptor_override_environment.yaml") as f:
+                override_environment = yaml.safe_load(f)
+
+            # Read DEVELOPMENT.md for instructions to create the wheels directory.
+            wheels_path = Path(__file__).parent.parent.parent.parent / "wheels"
+            if not wheels_path.exists() and wheels_path.is_dir():
```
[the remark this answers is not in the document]
leon-li-inspire: Good catch — fixed in 2ac45ba. Changed the guard to `if not wheels_path.exists() or not wheels_path.is_dir():` so the user-facing RuntimeError fires instead of a bare FileNotFoundError from `os.listdir`. Pre-existing bug carried over from `maya_render_submitter._get_job_template`. Added a regression test (`test_missing_wheels_dir_raises_runtimeerror`) asserting the missing-dir case raises RuntimeError.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,806 @@
...
+
+def get_default_job_template() -> dict[str, Any]:
+    """Returns the default job template for Maya render submissions."""
+    with open(Path(__file__).parent / "default_maya_job_template.yaml") as fh:
+        return yaml.safe_load(fh)
+
+
+def _get_render_layer_data() -> list[RenderLayerData]:
+    # Create a dictionary for the layers, and accumulate data about each layer
+    print(f"_get_render_layer_data - get_all_renderable_render_layer_names {time.time()}")
```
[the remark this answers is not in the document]
leon-li-inspire: Done in 42298ff — the four `print(f"... {time.time()}")` profiling statements in `_get_render_layer_data` are now `logger.debug(...)` calls (and the unused `time` import was dropped), so headless/programmatic use of `MayaSubmitter` no longer spams stdout.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,806 @@
...
+    scene changes.
+    """
+
+    def __init__(self, scene_context: Optional[MayaSceneContext] = None) -> None:
+        # Allow injecting a pre-built context (used by the GUI, which builds one
+        # for its own camera-list population, and by tests).
+        self._scene_context = scene_context
+
+    @property
+    def scene_context(self) -> MayaSceneContext:
```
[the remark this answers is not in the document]
leon-li-inspire: Verified + addressed in 42298ff. Against the released deadline 0.60.2, `BaseSubmitter.__init__ is object.__init__` — the base is a pure ABC (MRO `BaseSubmitter -> ABC -> object`, only the abstract methods + `get_submission_context`), so nothing was left unset. Added `super().__init__()` anyway so we stay correct if the base later initializes shared state.

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -856,10 +129,15 @@ def on_create_job_bundle_callback(
...
+    submitter = MayaSubmitter()
+    context = submitter.scene_context
 
-    # Tell the settings tab the selectable cameras
+    # Tell the settings tab the selectable cameras. The context lists already
+    # begin with ALL_CAMERAS (see create_submission_context); this preserves the
+    # pre-refactor assignment exactly.
     render_settings.current_layer_selectable_cameras = [ALL_CAMERAS] + sorted(
         context.current_layer_selectable_cameras
     )
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in c7715b2 — dropped the redundant prepend; the callback now assigns `context.current_layer_selectable_cameras` / `context.all_layer_selectable_cameras` directly, since `create_scene_context()` already returns them sorted with `ALL_CAMERAS` first. No more duplicate sentinel in the settings-tab dropdown.

test/unit/deadline_submitter_for_maya/scripts/test_submitter.py
```diff
@@ -231,3 +231,192 @@ def test_cameras_sorted_alphabetically(self):
             "middle_cam",
             "zebra_cam",
         ]
+
+
+class TestGuiSubmissionGlue:
+    """Covers the GUI-boundary glue introduced by the SubmitterAPI refactor:
```
leon-li-inspire: nit: SubmitterAPI -> BaseSubmitter

leon-li-inspire: Fixed in fa78e15 — docstring now says "BaseSubmitter refactor".

justinsaws:
The PR description said that a submission hasn't been done with this change yet. Has this been completed?

justinsaws:
Can you clean up the git history so that there is only one commit?

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,809 @@
...
+        """The cached scene context, built on first access."""
+        if self._scene_context is None:
+            self._scene_context = create_scene_context()
+        return self._scene_context
+
+    def reset_scene_context(self) -> None:
+        """Drop the cached scene context so the next access rescans the scene."""
+        self._scene_context = None
+
+    def get_settings(self) -> MayaSubmitterSettings:
```
justinsaws, line 218: `get_settings()` is new and has no direct test coverage, yet it's the default source of `settings` inside the inherited `get_submission_context()` — so it's the linchpin of the entire unified/AYON submission path.
  Please add a test that exercises `get_settings()` and ideally a round-trip (`get_settings()` → `get_job_template()` → `get_parameter_values()`) to lock down the headless contract.

src/deadline/maya_submitter/submitter.py
```diff
@@ -0,0 +1,809 @@
...
+        first_frame_range = submit_render_layers[0].frame_range
+        per_layer_frames_parameters = not settings.override_frame_range and any(
+            layer.frame_range != first_frame_range for layer in submit_render_layers
+        )
+
+        # If there are multiple frame ranges and we're not overriding the range,
+        # then we create per-layer Frames parameters.
+        if per_layer_frames_parameters:
+            for layer_data in submit_render_layers:
+                layer_data.frames_parameter_name = f"{layer_data.display_name}Frames"
```
justinsaws, line 401: Per-layer parameter names are keyed on `display_name` (`f"{layer_data.display_name}Frames"`, and the same below for output-prefix/width/height). Two render layers sharing a display name would produce duplicate
  parameter definitions in the job template. This is a pre-existing assumption, but the multi-layer path is now the headline feature — worth a defensive check (or at least a comment documenting the
  display-name-uniqueness assumption).

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -856,16 +129,18 @@ def on_create_job_bundle_callback(
...
-        context.current_layer_selectable_cameras
-    )
-    render_settings.all_layer_selectable_cameras = [
-        ALL_CAMERAS
-    ] + context.all_layer_selectable_cameras
+    # Tell the settings tab the selectable cameras. create_scene_context()
+    # already returns these lists sorted with ALL_CAMERAS prepended, so assign
+    # them directly — prepending ALL_CAMERAS again would duplicate the sentinel
+    # in the dropdown.
+    render_settings.current_layer_selectable_cameras = context.current_layer_selectable_cameras
```
justinsaws: `render_settings` (from `_set_render_setting()` above) gets the camera lists assigned here but is never used again in this callback — the builders below run off `settings`/`maya_settings`. So this camera-list
  population is a dead assignment, and the comment's claim that query-once spans "the camera-list population below and both builder calls" is a bit misleading (the real benefit is just the two builders sharing the
  cached context). Either drop the dead assignment or adjust the comment.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Added in bf78e6b — new TestGetSettings covers get_settings() (reads-scene + unsaved-scene job_name fallback) plus a headless round-trip get_settings() -> get_job_template() -> get_parameter_values() asserting name/ProjectPath/OutputFilePath flow through.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Documented in bf78e6b — added a comment in _prepare_render_layers_for_submission noting the per-layer parameter names key on display_name and assume it's unique across submitted layers (renderSetup enforces unique layer names in practice). Left it as a documented assumption rather than a hard guard, per your 'or at least a comment' suggestion.

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: Fixed in bf78e6b — dropped the dead render_settings camera-list assignment (render_settings was never read again in the callback; the dialog is populated in show_maya_render_submitter). Corrected the comment: the query-once benefit is just the two builders sharing one cached scene context.

leon-li-inspire:
Yes — an end-to-end submission has been completed with this change (GUI publish → MayaSubmitter → CreateJob rendered successfully on a live farm). I'll update the PR description to drop the 'pending' note.

leon-li-inspire:
Done — squashed the branch to a single commit (`feat: implement MayaSubmitter on the unified BaseSubmitter`) and force-pushed. The resulting tree is byte-identical to the prior reviewed state (no content change); CI is re-running on the new commit.

src/deadline/maya_submitter/submitter.py
leon-li-inspire: Switched to a defensive check (e07db07) per your preference: `_prepare_render_layers_for_submission` now raises `DeadlineOperationError` listing any duplicated display names before building per-layer parameters, so a collision fails loudly instead of silently dropping a layer's frame range / prefix / resolution. Added test `test_prepare_duplicate_display_names_raise`.

justinsaws (approved):

andychoquette (approved):

PR #437, opened 2026-07-07
feat: Run pre-GUI hooks in the Maya render submitter

leon-li-inspire:
Fixes: *N/A — tracked internally; will link a GitHub issue if one is opened.*

> **Depends on `deadline-cloud >= 0.60.1`** (published), which ships the pre-GUI hooks API (`deadline.client.ui.pre_gui_hooks`: `run_pre_gui_hooks` / `PreGuiHookContext` / `apply_pre_gui_output`). The dependency floor is pinned accordingly. Remaining before merge: post the integ + Job Bundle Output Test results (require a Maya environment).

Pre-GUI submission hooks did not fire in DCC submitters. Studios can define pre-GUI hooks (via `DEADLINE_HOOKS_DIR`) to pre-populate submission fields — job name, description, and queue/job parameters — before the submit dialog opens, but the Maya submitter never invoked them. This PR is the **Maya adoption** of the shared pre-GUI hooks mechanism; the entry point (`run_pre_gui_hooks`) is provided by `deadline-cloud` so every DCC submitter can call it consistently.

In `show_maya_render_submitter`, immediately before constructing `SubmitJobToDeadlineDialog`:

- Build a `shared_parameter_values` dict (the `RezPackages` / `CondaPackages` that were previously an inline literal on the dialog).
- Call `run_pre_gui_hooks(PreGuiHookContext(...))` with:
  - `bundle_dir=None` — Maya has no on-disk job bundle at pre-GUI time, so hooks are sourced from `DEADLINE_HOOKS_DIR` only (gated by `settings.allow_environment_hooks`).
  - `submitter_name="maya"`, `job_name=render_settings.name`, and the current `parameters` passed in so a hook can read/override them.
  - A `confirm_callback` that follows the auto-accept policy: `None` (no prompt) when `settings.auto_accept` is set, otherwise `qt_hook_confirmation(parent)` for the standard confirmation dialog.
- Map the merged hook output back with deadline-cloud's generic `apply_pre_gui_output`. `RenderSubmitterUISettings` has no `.parameters` list (unlike the standalone submitter's `JobBundleSettings`), so `apply_pre_gui_output` writes `name` / `description` onto the settings object and routes every hook `parameter` into `shared_parameter_values` — exactly how a DCC seeds the dialog.

The pre-GUI hooks symbols are imported lazily inside `show_maya_render_submitter` (mirroring the existing lazy `qtpy` import there) so the module — and every unit test that collects it — imports cleanly, and the hooks are only required when the submitter is actually launched.

Also pins the `deadline` dependency floor to `>= 0.60.1, < 0.61` (the first published release that ships the pre-GUI hooks API).

- **No behavior change when no hooks are configured.** With no `DEADLINE_HOOKS_DIR` env hooks, `run_pre_gui_hooks` returns an empty dict, `apply_pre_gui_output` is a no-op, and the dialog receives exactly the same `RezPackages` / `CondaPackages` as before.
- The only new behavior fires when environment hooks are present and enabled: hooks run (with a confirmation prompt unless `auto_accept`), and their output pre-populates the dialog.

- `hatch run lint` (black + ruff) — clean on both changed files.
- Unit tests in `test/unit/deadline_submitter_for_maya/test_pre_gui_hooks.py` (below) — now run against `deadline-cloud 0.60.1` in CI (all Python matrix legs green).
- Manual: full `show_maya_render_submitter` requires a running Maya (integration suite); the mapping is unit-tested headless under the package's maya/qtpy `sys.modules` stubs.

#### Integ test result

_Pending — to be run against `deadline-cloud 0.60.1` in a Maya environment; results will be posted here._

#### Installer test result

N/A — `installer/` not modified.

_Pending — to be run in a Maya environment; `job_bundle_output_tests/test-job-bundle-results.txt` will be posted here._

Unit tests added: `test/unit/deadline_submitter_for_maya/test_pre_gui_hooks.py` — 4 tests verifying that deadline-cloud's `apply_pre_gui_output` routes hook output correctly against Maya's `RenderSubmitterUISettings` (name/description onto settings, hook parameters into shared values with override-on-collision, empty-output no-op, partial output). Uses `importorskip` so it skips cleanly on deadline-cloud releases that predate the API.

- Docstring/comment added at the inline hook-invocation block explaining the env-only sourcing, the auto-accept confirmation policy, and why the import is lazy.
- No README/DEVELOPMENT changes needed — this uses the existing `DEADLINE_HOOKS_DIR` / `allow_environment_hooks` mechanism documented on the deadline-cloud side.

- [ ] Yes
- [x] No

No. Public interfaces are unchanged, and behavior with no configured hooks is identical to before.

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,9 +30,6 @@ classifiers = [
 ]
 
 dependencies = [
-    # 0.60.1 is the first release to ship deadline.client.ui.pre_gui_hooks
-    # (run_pre_gui_hooks / PreGuiHookContext / qt_hook_confirmation / apply_pre_gui_output), which
-    # the pre-GUI hook integration in maya_render_submitter.show_maya_render_submitter requires.
     "deadline >= 0.60.1,< 0.61",
     "openjd-adaptor-runtime == 0.9.*",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,10 +30,14 @@ classifiers = [
 ]
 
 dependencies = [
-    # run_pre_gui_hooks / PreGuiHookContext (deadline.client.ui.pre_gui_hooks) are required for
-    # the pre-GUI hook integration below. That module was added after the 0.60.0 tag, so it ships
-    # in the next minor release, 0.61.0. Re-confirm against the published release before merging.
-    "deadline >= 0.61.0,< 0.62",
+    # TODO: bump this floor to the deadline-cloud release that ships
+    # deadline.client.ui.pre_gui_hooks (run_pre_gui_hooks / PreGuiHookContext /
+    # apply_pre_gui_output) before marking this PR ready. That module was added after the
+    # 0.60.0 tag and is not published yet, so the floor is held at the latest released version
+    # (0.60.0) to keep dependency resolution working. The pre-GUI hooks import is done lazily
+    # in maya_render_submitter.show_maya_render_submitter, so this module (and the unit tests)
+    # import cleanly on 0.60.0; the hooks are only required when the submitter is launched.
+    "deadline >= 0.60.0,< 0.61",
     "openjd-adaptor-runtime == 0.9.*",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,14 +30,10 @@ classifiers = [
 ]
 
 dependencies = [
-    # TODO: bump this floor to the deadline-cloud release that ships
-    # deadline.client.ui.pre_gui_hooks (run_pre_gui_hooks / PreGuiHookContext /
-    # apply_pre_gui_output) before marking this PR ready. That module was added after the
-    # 0.60.0 tag and is not published yet, so the floor is held at the latest released version
-    # (0.60.0) to keep dependency resolution working. The pre-GUI hooks import is done lazily
-    # in maya_render_submitter.show_maya_render_submitter, so this module (and the unit tests)
-    # import cleanly on 0.60.0; the hooks are only required when the submitter is launched.
-    "deadline >= 0.60.0,< 0.61",
+    # 0.60.1 is the first release to ship deadline.client.ui.pre_gui_hooks
+    # (run_pre_gui_hooks / PreGuiHookContext / apply_pre_gui_output), which the pre-GUI hook
+    # integration in maya_render_submitter.show_maya_render_submitter requires.
+    "deadline >= 0.60.1,< 0.61",
     "openjd-adaptor-runtime == 0.9.*",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,9 +30,10 @@ classifiers = [
 ]
 
 dependencies = [
-    # run_pre_gui_hooks / PreGuiHookContext are required for the pre-GUI hook integration
-    # below; they ship in deadline-cloud 0.60.x. Confirm the exact release before merging.
-    "deadline >= 0.60.0,< 0.61",
+    # run_pre_gui_hooks / PreGuiHookContext (deadline.client.ui.pre_gui_hooks) are required for
+    # the pre-GUI hook integration below. That module was added after the 0.60.0 tag, so it ships
+    # in the next minor release, 0.61.0. Re-confirm against the published release before merging.
+    "deadline >= 0.61.0,< 0.62",
     "openjd-adaptor-runtime == 0.9.*",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,9 @@ classifiers = [
 ]
 
 dependencies = [
-    "deadline >= 0.59.0,< 0.60",
+    # run_pre_gui_hooks / PreGuiHookContext are required for the pre-GUI hook integration
+    # below; they ship in deadline-cloud 0.60.x. Confirm the exact release before merging.
+    "deadline >= 0.60.0,< 0.61",
     "openjd-adaptor-runtime == 0.9.*",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,8 +31,8 @@ classifiers = [
 
 dependencies = [
     # 0.60.1 is the first release to ship deadline.client.ui.pre_gui_hooks
-    # (run_pre_gui_hooks / PreGuiHookContext / apply_pre_gui_output), which the pre-GUI hook
-    # integration in maya_render_submitter.show_maya_render_submitter requires.
+    # (run_pre_gui_hooks / PreGuiHookContext / qt_hook_confirmation / apply_pre_gui_output), which
+    # the pre-GUI hook integration in maya_render_submitter.show_maya_render_submitter requires.
     "deadline >= 0.60.1,< 0.61",
     "openjd-adaptor-runtime == 0.9.*",
 ]
--- a/src/deadline/maya_submitter/maya_render_submitter.py
+++ b/src/deadline/maya_submitter/maya_render_submitter.py
@@ -22,11 +22,6 @@ from deadline.client.ui.dialogs.submit_job_to_deadline_dialog import (  # pylint
     SubmitJobToDeadlineDialog,
     JobBundlePurpose,
 )
-from deadline.client.ui.pre_gui_hooks import (  # pylint: disable=import-error
-    PreGuiHookContext,
-    qt_hook_confirmation,
-    run_pre_gui_hooks,
-)
 from deadline.client.exceptions import DeadlineOperationError
 from qtpy.QtCore import Qt  # type: ignore
 
@@ -1059,6 +1054,16 @@ def show_maya_render_submitter(
     # on-disk job bundle at this point, so hooks are sourced from DEADLINE_HOOKS_DIR only
     # (bundle_dir=None), gated by settings.allow_environment_hooks. The confirmation prompt is
     # skipped when auto_accept is set; otherwise the standard dialog is shown.
+    #
+    # Imported lazily (like qtpy above) rather than at module top: this ships in deadline-cloud
+    # 0.60+, and a top-level import would break importing this module — and every unit test that
+    # collects it — against older deadline-cloud releases.
+    from deadline.client.ui.pre_gui_hooks import (  # pylint: disable=import-error
+        PreGuiHookContext,
+        qt_hook_confirmation,
+        run_pre_gui_hooks,
+    )
+
     confirm_callback = (
         None if str2bool(get_setting("settings.auto_accept")) else qt_hook_confirmation(parent)
     )
--- a/src/deadline/maya_submitter/maya_render_submitter.py
+++ b/src/deadline/maya_submitter/maya_render_submitter.py
@@ -22,7 +22,7 @@ from deadline.client.ui.dialogs.submit_job_to_deadline_dialog import (  # pylint
     SubmitJobToDeadlineDialog,
     JobBundlePurpose,
 )
-from deadline.client.exceptions import DeadlineOperationError
+from deadline.client.exceptions import DeadlineOperationCanceled, DeadlineOperationError
 from qtpy.QtCore import Qt  # type: ignore
 
 from . import Animation, Scene  # type: ignore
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -30,7 +30,9 @@ classifiers = [
 ]
 
 dependencies = [
-    "deadline >= 0.59.0,< 0.60",
+    # run_pre_gui_hooks / PreGuiHookContext are required for the pre-GUI hook integration
+    # below; they ship in deadline-cloud 0.60.x. Confirm the exact release before merging.
+    "deadline >= 0.60.0,< 0.61",
```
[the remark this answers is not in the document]
leon-li-inspire: Good catch — verified, and the pin was wrong in both directions. Findings against the deadline-cloud source:

- `deadline.client.ui.pre_gui_hooks` (with `run_pre_gui_hooks`, `PreGuiHookContext`, `qt_hook_confirmation`) is **not present at the `0.60.0` tag** — it was added afterward in `feat: add Qt-free run_pre_gui_hooks for DCC submitters`. `0.60.0` is currently the latest published tag.
- The commits since `0.60.0` include `feat`s, so semantic-release will cut a **minor** bump → **`0.61.0`**. My old upper bound `< 0.61` would have *excluded* the very release that contains the module.
- `str2bool` (from `deadline.client.config`) does exist at `0.60.0`, so that import was fine.

Fixed in `5a0adce`: floor bumped to `deadline >= 0.61.0, < 0.62`, comment updated to explain the post-0.60.0 sourcing. This remains a hard merge prerequisite — the PR stays draft until deadline-cloud `0.61.0` is published, and I'll re-confirm the symbols against the actual PyPI release before marking ready.

leon-li-inspire:
### Re-verified against the latest deadline-cloud PR #1255

deadline-cloud PR #1255 was updated (commits `5876bc5` *"declare public API via `__all__` and drop internal type from signatures"* and `6923583` *"expose HookManager publicly for pre-GUI confirm_callback typing"*). I re-checked this Maya PR against it — **no changes needed here.** Details:

- **The changes are API-hygiene only.** `pre_gui_hooks` now declares `__all__ = ["PreGuiHookContext", "apply_pre_gui_output", "qt_hook_confirmation", "run_pre_gui_hooks"]` — which is exactly the set this submitter imports, so we're on the officially-declared public API. `HookManager` was moved to a public module and used in internal type annotations only.
- **All four symbols we call are unchanged** in signature: `PreGuiHookContext`, `run_pre_gui_hooks(context, *, confirm_callback=...)`, `qt_hook_confirmation(parent)`, and `apply_pre_gui_output(pre_gui_output, initial_settings, initial_shared_parameter_values, cli_provided_param_names=None)`.
- This submitter imports **only** from `deadline.client.ui.pre_gui_hooks`; it never references `HookManager` / `HookMetadata` / `job_bundle._hooks`, so the internal reorganization is invisible here.
- Ran the mapping tests against the updated core: **4/4 pass**.

No commit required for this. The one remaining pre-merge item is unchanged: the `deadline` floor is held at `>= 0.60.0` (latest published) with a TODO to bump it to the release that ships `deadline.client.ui.pre_gui_hooks` once that release is out.

pyproject.toml
```diff
@@ -30,7 +30,10 @@ classifiers = [
 ]
 
 dependencies = [
-    "deadline >= 0.59.0,< 0.60",
+    # 0.60.1 is the first release to ship deadline.client.ui.pre_gui_hooks
+    # (run_pre_gui_hooks / PreGuiHookContext / apply_pre_gui_output), which the pre-GUI hook
+    # integration in maya_render_submitter.show_maya_render_submitter requires.
+    "deadline >= 0.60.1,< 0.61",
```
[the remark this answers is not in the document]
leon-li-inspire: Reconciled in `60812b1`. All three version statements now agree on **0.60.1**: verified against the *published* `deadline 0.60.1` wheel, `deadline/client/ui/pre_gui_hooks.py` is present and exports `PreGuiHookContext`, `run_pre_gui_hooks`, `qt_hook_confirmation`, and `apply_pre_gui_output` (plus `__all__`). It was a patch bump (0.60.1), not the 0.61.0 I'd earlier guessed. The pyproject floor is `>= 0.60.1, < 0.61`, and the code comment + test docstring now both say `0.60.1+` — so `< 0.61` correctly includes the release that ships the API. CI installs 0.60.1 and the `importorskip` test now actually runs (all Python legs green).

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -1022,13 +1022,46 @@ def update_progress(message):
...
+    # (bundle_dir=None), gated by settings.allow_environment_hooks. The confirmation prompt is
+    # skipped when auto_accept is set; otherwise the standard dialog is shown.
+    #
+    # Imported lazily (like qtpy above) rather than at module top: this ships in deadline-cloud
+    # 0.60.1+, and a top-level import would break importing this module — and every unit test that
+    # collects it — against older deadline-cloud releases.
+    from deadline.client.ui.pre_gui_hooks import (  # pylint: disable=import-error
+        PreGuiHookContext,
+        apply_pre_gui_output,
+        qt_hook_confirmation,
```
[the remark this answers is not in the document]
leon-li-inspire: Confirmed: `qt_hook_confirmation` **is** exported as of 0.60.1. Verified against the published `deadline-0.60.1-py3-none-any.whl` — `deadline/client/ui/pre_gui_hooks.py` defines `qt_hook_confirmation` (line 174) and lists it in `__all__` alongside the other three symbols, so all four the submitter imports ship in the same 0.60.1 release. No runtime `ImportError` risk on the allowed `0.60.x` range.

Addressed in `1cb3839`: added `qt_hook_confirmation` to the enumerated symbols in the `pyproject.toml` dependency note so it matches the import.

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -1022,13 +1022,46 @@ def update_progress(message):
...
+
+    # Run pre-GUI hooks so studios can pre-populate dialog fields before it opens. Maya has no
+    # on-disk job bundle at this point, so hooks are sourced from DEADLINE_HOOKS_DIR only
+    # (bundle_dir=None), gated by settings.allow_environment_hooks. The confirmation prompt is
+    # skipped when auto_accept is set; otherwise the standard dialog is shown.
+    #
+    # Imported lazily (like qtpy above) rather than at module top: this ships in deadline-cloud
+    # 0.60.1+, and a top-level import would break importing this module — and every unit test that
+    # collects it — against older deadline-cloud releases.
+    from deadline.client.ui.pre_gui_hooks import (  # pylint: disable=import-error
```
crowecawcaw: Is this pylint disable from before the release?

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -1022,13 +1022,46 @@ def update_progress(message):
...
+        PreGuiHookContext,
+        apply_pre_gui_output,
+        qt_hook_confirmation,
+        run_pre_gui_hooks,
+    )
+
+    confirm_callback = (
+        None if str2bool(get_setting("settings.auto_accept")) else qt_hook_confirmation(parent)
```
crowecawcaw: Note: we should update get_setting to return bools for settings that are boolean instead of having callers sort it out

pyproject.toml
```diff
@@ -30,7 +30,10 @@ classifiers = [
 ]
 
 dependencies = [
-    "deadline >= 0.59.0,< 0.60",
+    # 0.60.1 is the first release to ship deadline.client.ui.pre_gui_hooks
+    # (run_pre_gui_hooks / PreGuiHookContext / qt_hook_confirmation / apply_pre_gui_output), which
+    # the pre-GUI hook integration in maya_render_submitter.show_maya_render_submitter requires.
```
crowecawcaw: Nit: don't need this comment

test/unit/deadline_submitter_for_maya/test_pre_gui_hooks.py
```diff
@@ -0,0 +1,94 @@
...
+
+import pytest
+
+# apply_pre_gui_output ships in deadline-cloud 0.60.1+; skip cleanly on older releases.
+pre_gui_hooks = pytest.importorskip("deadline.client.ui.pre_gui_hooks")
+apply_pre_gui_output = pre_gui_hooks.apply_pre_gui_output
+
+from deadline.maya_submitter.data_classes import (  # noqa: E402  (after importorskip)
+    RenderSubmitterUISettings,
+)
```
crowecawcaw: Nit: don't need this skip since we're updating requirements and release is out

crowecawcaw (changes requested):
Looks good but need cleanup from defense coding from before the release

src/deadline/maya_submitter/maya_render_submitter.py
```diff
@@ -1022,13 +1022,46 @@ def update_progress(message):
...
+    )
```
crowecawcaw: Do we need a test that this confirmation dialog fires when auto_accept is not enabled?

Also is there a constant in `deadline` lib we can use for this config name?

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: Good catch — removed in `68eb39a`. pylint isn't part of the lint stack (this repo runs ruff + black + mypy), and `deadline` is an installed dependency, so the `import-error` disable was inert. (The two identical disables at the top of the file predate this PR; left them alone to keep the change focused.)

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: Agreed that's a nice improvement, but it's in the core `deadline` library's `get_setting`, not this repo. This call follows the existing convention deadline-cloud itself uses for the same setting (`str2bool(get_setting("settings.auto_accept"))` in `job_bundle_submitter.py`), so I've kept it consistent here. Happy to file a follow-up against deadline-cloud to have `get_setting` type-coerce booleans — leaving this thread for you to resolve/track as you see fit.

pyproject.toml
leon-li-inspire: Removed in `68eb39a`.

test/unit/deadline_submitter_for_maya/test_pre_gui_hooks.py
leon-li-inspire: Removed in `68eb39a` — the test now imports `apply_pre_gui_output` directly. Since the floor is `deadline >= 0.60.1` (which ships the module) the `importorskip` guard (and its `noqa: E402`) are no longer needed.

src/deadline/maya_submitter/maya_render_submitter.py
leon-li-inspire: Both addressed in `68eb39a`:
- **Test:** extracted the auto_accept selection into `_pre_gui_hook_confirm_callback(parent)` and added two unit tests — one asserting it returns `None` when `settings.auto_accept` is enabled (no prompt), and one asserting it returns `qt_hook_confirmation(parent)` otherwise.
- **Config-name constant:** there isn't an exported one in the `deadline` lib — the setting is a key in the internal `config_file.SETTINGS` dict, and deadline-cloud's own `job_bundle_submitter.py` uses the string literal `"settings.auto_accept"` directly. Kept the literal to match that precedent; if a public constant is added upstream we can switch to it.

leon-li-inspire: You're right — the first version mocked out `qt_hook_confirmation` and only asserted a non-`None` callback was selected, so it never verified the dialog actually fires. Strengthened in `f958889`:

`test_confirmation_dialog_fires_when_auto_accept_disabled` now leaves `qt_hook_confirmation` real, **invokes** the returned callback, and asserts:
- `QMessageBox.question` is called exactly once (the prompt actually fires),
- it is parented to the window passed into the submitter (`call_args[0][0] == "mainwin"`),
- a `Yes` reply maps to `True` (proceed).

So it exercises the prompt firing end-to-end through `_pre_gui_hook_confirm_callback` → `qt_hook_confirmation`, rather than just branch selection. (The dialog's message-formatting internals live in — and are tested by — deadline-cloud's `qt_hook_confirmation`.)

crowecawcaw (approved):

jericht (approved):

PR #453, opened 2026-07-26
Fix Renderers that store their output V-Ray ignores scene's vraySettings.fileNamePrefix; submitter falls back to <Scene>/<Layer>

dzavada:
What was the problem/requirement? (What/Why)
V-Ray ignores scene's vraySettings.fileNamePrefix; submitter falls back to / default 443

What was the solution? (How)
Renderers that store their output file name prefix somewhere other than
defaultRenderGlobals.imageFilePrefix. V-Ray's Render Settings Common tab
writes the "File Name Prefix" field to vraySettings.fileNamePrefix and does
not update the legacy defaultRenderGlobals attribute, so reading the default
attribute for a V-Ray scene picks up a stale/empty value and the submitted
job ignores the prefix the user actually configured.
What is the impact of this change?
N/A

How was this change tested?
I submitted jobs through Maya and tested that is in fact was following the correct naming prefix.

Integ test result
delete text starting here
If src/ was modified or a file was added/removed, then update the integ tests and post the test results below. See
DEVELOPMENT.md on additional instructions on running some, or all integ tests.
delete text ending here

Installer test result
ADAPTOR_OUTPUT: STDOUT: Performing action: {"name": "output_file_prefix", "args": {"output_file_prefix": "Mavic/%s/%l/%l_%v"}}

Did you run the "Job Bundle Output Tests"? If not, why not? If so, paste the test results here.
No

Was this change documented?
No

Did you modify schema files?
 Yes
[ X] No
If yes, Did you bump the integration_data_interface_version in src/deadline/maya_adaptor/MayaAdaptor/adaptor.py and update the test constants?

 Yes
[ X] No
Is this a breaking change?
No


dzavada (commented):
Fix formatting in renderers.py

PR #463, opened 2026-08-19
fix(ci): Allow single-platform Maya installers

andrew-fenton:

`pipeline/setup-runner.py` required every Maya version to declare installers for
both platforms, so a version available on only one platform could not be added to
the `integ-ci` matrix without failing the other platform's leg.

Installer and checksum maps are now per-platform dicts, and a version with no
installer for the current platform is skipped instead of failing the run.

**These are the failure/skip cases that can happen under this change:**
| Situation | Result |
|---|---|
| Version not in `MAYA_YEAR_TO_CONFIG` | error, exit 1 |
| Version declared, `installer` empty | error, exit 1 |
| Version declared, absent on this platform only | `SKIP:`, continue |
| No requested version installable here | message, exit 0 |

CI-only; `pipeline/` is not shipped with the pip package.

`hatch run test` — 13 new unit tests covering the four cases above and confirming checksum enforcement is unchanged.

#### Integ test result

Linux and Windows both pass against this PR's head `b7f221c` (CodeBuild source version `pr/463`): [run 32317369890](https://github.com/aws-deadline/deadline-cloud-for-maya/actions/runs/32317369890)

#### Installer test result

Not run. `installer/` is unmodified and no files were added to or removed from `src/`.

No. The submitter and adaptor are untouched, so no job bundle output can change.

The module docstring now states the per-platform installer rule. No user-facing docs are affected.

- [ ] Yes
- [x] No

No.

----

### DIFF
--- a/pipeline/setup-runner.py
+++ b/pipeline/setup-runner.py
@@ -2,7 +2,13 @@
 #!/usr/bin/env python3
 """Setup runner for Maya integration tests in CodeBuild.
 
-Supports Linux and Windows with Maya 2025 and 2026.
+Supports Linux and Windows for the Maya versions in MAYA_YEAR_TO_CONFIG below.
+
+A Maya version only needs an installer for the platforms it is available on. A
+configured version with no installer for the current platform is skipped with a
+warning instead of failing the run, so a Maya version can be added to the test
+matrix as soon as one platform's installer is available. Requesting a version
+that is not in MAYA_YEAR_TO_CONFIG at all is still an error.
 
 Installs the Arnold (MtoA), V-Ray, and Redshift renderers into each Maya
 version so the renderer-specific integ tests can run.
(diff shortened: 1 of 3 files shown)


andrew-fenton:
Closing unimplemented. The 2027 rollout no longer needs this.

This PR let a Maya version declare an installer for only one platform. In practice Maya 2027 ships installers for both, and the actual blocker was the renderers: V-Ray and Redshift have no 2027 builds, and `_install_vray_linux` / `_install_mtoa_linux` exit 1 on an unconfigured version while their Windows counterparts warn and skip.

That is better handled with a hatch matrix override on the 2027 cell, which installs only MtoA and deselects the two renderer tests, so the Linux `sys.exit(1)` is never reached and no leniency is required.

Both review findings here were artifacts of this change rather than pre-existing issues, and closing retires them:

- `maya_checksum()` returning `""` weakened a missing checksum from `KeyError` to a warning.
- `sys.exit(0)` made a skipped setup indistinguishable from a successful one.

One finding is worth pursuing separately: on mainline today, `verify_checksum()` prints `WARNING: No checksum configured ... skipping verification` and returns `True`, so a forgotten checksum means an installer is downloaded and executed as root unverified. Making that a hard error applies to every installer, including the renderers, and is independent of anything in this PR.

PR #476, opened 2026-09-09
fix(ci): Repair the Linux Maya install in setup-runner

andrew-fenton:

In Progress


andrew-fenton:
Superseded by #479. Same change, squashed to a single commit and rebased onto mainline now that #477 has merged, with a description that matches what the change actually became — this thread accumulated stale comments as the diagnosis moved on (the libssl compat symlinks and `libGLU` claims here turned out to be wrong, and the SONAME, mayapy dispatcher and `MAYA_VERSION` work all landed after the original description was written).

PR #477, opened 2026-09-10
ci: Sweep orphaned Maya processes before and after integ tests

andrew-fenton:
## Problem

Linux integ builds have failed continuously since 2026-09-03:

```
RuntimeError: Error encountered when initializing Maya - Please check for
sufficient disk space and necessary write permissions of MAYA_APP_DIR.
```

The message is misleading. Disk and `MAYA_APP_DIR` are fine; the real cause is a refused Autodesk license checkout.

A suite that completes normally releases its licenses, so this is not a per-run leak. Orphans appear when a run is killed before its subprocess chain is torn down, such as a cancelled or timed-out build. They then keep their license session alive indefinitely, accumulating across builds until every later checkout is refused.

The failing host had **22 orphaned `mayapy` processes spanning 10 previous builds**, all Maya 2027 while CI tests only 2025 and 2026, so they came from aborted development runs rather than CI itself.

## Fix

Sweep orphaned Maya processes in `pipeline/run-integ-tests.py`, before and after the suite. Adds `psutil` to the integ test requirements.

Also closes a second leak path found in review: the `initialize_maya` teardown called `qt_application.shutdown()` before `maya.standalone.uninitialize()` with no `try/finally`, so a Qt shutdown error leaked the license from an otherwise normal run.

## Testing

Verified end to end on the fleet: the sweep terminated 22 orphans, then `10 passed` on both Maya versions, where the previous run had 4 failed and 5 errored.

Unit tests in `test/unit/test_run_integ_tests.py` cover the two judgement calls a later change could silently reverse: a cleanup failure never failing the build, and the marker set not matching processes that merely mention Maya. The second is mutation-tested; adding `"Render"` to the markers fails it.

Deliberately not unit-tested: psutil and OS behaviour. A fake psutil can only return what we assume psutil returns, so it cannot validate that assumption.

## Appendix

<details>
<summary><b>Design notes</b> (why the runner, why two sweeps, why these markers)</summary>

**Why the runner, not a fixture or `conftest.py`:** pytest itself runs under `mayapy`, so a sweep inside it would match and kill its own interpreter. The runner is plain Python that launches `mayapy` as a subprocess, so it can terminate Maya processes safely. It also runs once, avoiding xdist worker duplication.

**Why both pre- and post-test:** post-test stops this build poisoning the next; pre-test recovers from builds that never reached their post-test sweep, such as a timeout or cancellation. Neither covers the other's case.

**Why CodeBuild-only:** `CODEBUILD_BUILD_ID` guards the sweep so a local `hatch run integ-ci:test` cannot kill a Maya the developer opened.

**Why these process markers:** every orphan observed was `mayapy`. The renderers run in-process as Maya plugins, so terminating `mayapy` releases the Maya, Arnold, V-Ray and Redshift checkouts together; there are no separate renderer processes to match. `maya.bin` and `maya.exe` are precautionary, and `maya.exe` covers the Windows integ project, which shares this script and the same reused-host behaviour.

**Why `psutil`:** the runner executes outside Maya's interpreter, where `psutil` was previously only available transitively.

**Why cleanup cannot fail the build:** errors during the sweep are caught and logged, and the test exit code is preserved.

</details>

<details>
<summary><b>Build log evidence</b></summary>

The adaptor names the real cause itself, immediately after the misleading error:

```
STDERR: RuntimeError: Error encountered when initializing Maya - Please check
        for sufficient disk space and necessary write permissions of MAYA_APP_DIR.
This error is typically associated with a licensing error when using MayaIO.
Free disc space: 49644M
MAYA_APP_DIR: None
ADSKFLEX_LICENSE_FILE: 2702@<endpoint-dns>:2701@<endpoint-dns>
```

49 GB free and `MAYA_APP_DIR` unset, so neither cause named in the message applies.

The sweep on the first green run, trimmed:

```
[license-cleanup] terminating 22 orphaned Maya process(es):
[license-cleanup]   pid=97349   .../mayaIO2027/bin/mayapy.bin -m openjd run /tmp/codebuild-ff222003-.../
[license-cleanup]   pid=135751  .../mayaIO2027/bin/mayapy.bin -m openjd run /tmp/codebuild-1326f261-.../
[license-cleanup]   pid=3252356 .../mayaIO2027/bin/mayapy.bin /tmp/codebuild-63995b18-.../
```

Distinct `codebuild-*` directories confirm the orphans came from many earlier builds, and the PID range (97,349 to 3,252,356) shows they accumulated over a long period.

Then, on both matrix versions:

```
================== 10 passed, 4 warnings in 275.94s (0:04:35) ==================
...
[license-cleanup] post-test: no orphaned Maya processes found
================== 10 passed, 4 warnings in 145.94s (0:02:25) ==================
```

The second run found nothing to clean, so a passing suite releases its own licenses.

</details>

<details>
<summary><b>Considered and deferred</b> (buildspec <code>post_build</code> sweep)</summary>

A sweep in the buildspec `post_build` phase would additionally cover builds that fail before reaching the test phase. Deferred because only 1 of the last 20 builds failed that early, it cannot help with cancelled builds either, and post-test cleanup already bounds accumulation to a single build.

Worth revisiting if orphans reappear having survived a build that failed during install or pre-build, if something other than pytest starts running Maya on this fleet, or if cancelled builds stop being rare.

</details>


pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,116 @@
...
+    for proc in still_alive:
+        _log(f"  pid={proc.pid} ignored SIGTERM; sending SIGKILL")
+        try:
+            proc.kill()
+        except (psutil.NoSuchProcess, psutil.AccessDenied) as exc:
+            _log(f"  pid={proc.pid} kill failed: {exc}")
+
 
 def main():
+    release_orphaned_maya_licenses("pre-test")
```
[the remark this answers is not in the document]
andrew-fenton: Fixed in 748561a. Both call sites now go through a `_cleanup(phase)` helper. Confirmed the failure mode you describe: `process_iter` and `wait_procs` sit outside the `NoSuchProcess`/`AccessDenied` handlers, so this would have failed the build before any test ran.

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,99 @@
...
+    signalling anything.
+    """
+    orphans = []
+    for proc in psutil.process_iter(["pid", "cmdline"]):
+        if proc.pid in protected_pids:
+            continue
+        try:
+            cmdline = " ".join(proc.info.get("cmdline") or [])
+        except (psutil.NoSuchProcess, psutil.AccessDenied):
+            continue
```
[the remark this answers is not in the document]
andrew-fenton: Correct, and fixed in 748561a. Verified against psutil 7.2.2: `process_iter` swallows `NoSuchProcess`, `as_dict` catches `(AccessDenied, ZombieProcess)` and substitutes `ad_value`, and `proc.info` is a plain dict, so neither exception could be raised.

The silent-drop consequence was the real problem. `find_orphaned_maya_processes` now returns `(orphans, unreadable_pids)` and the caller logs the count. Measured on a dev machine, 431 of 1138 processes returned `cmdline=None`, so this is not a rare path.

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,99 @@
...
+    for proc in still_alive:
+        print(f"[license-cleanup]   pid={proc.pid} ignored SIGTERM; sending SIGKILL")
+        try:
+            proc.kill()
+        except (psutil.NoSuchProcess, psutil.AccessDenied) as exc:
+            print(f"[license-cleanup]   pid={proc.pid} kill failed: {exc}")
+
 
 def main():
+    release_orphaned_maya_licenses()
```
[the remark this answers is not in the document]
andrew-fenton: Two parts.

The teardown finding is right and is fixed in 748561a: if `qt_application.shutdown()` raised, `uninitialize()` never ran and the license leaked from an otherwise normal run. Now in a `try/finally`.

The "pre-run sweep only" part is out of date; this comment predates the post-test sweep, which was added in an earlier commit. Both sweeps now run, and they cover different cases: post-test stops this build poisoning the next, pre-test recovers from builds killed before reaching post-test.

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,99 @@
...
 
+# Command-line fragments identifying a process that can hold an Autodesk
+# (FlexLM) license checkout.
+_MAYA_PROCESS_MARKERS = ("mayapy", "maya_client.py", "maya.bin", "maya.exe")
+
+
+def find_orphaned_maya_processes(psutil, protected_pids):
+    """Return [(process, cmdline)] for Maya processes not in protected_pids.
+
+    Separated from termination so the matching logic can be exercised without
```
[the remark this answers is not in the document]
andrew-fenton: Added in 748561a, deliberately narrow: four tests covering the two judgement calls a later change could silently reverse, rather than mocked assertions about psutil behaviour that a fake cannot validate.

The guardrail is mutation-tested. Adding `"Render"` to `_MAYA_PROCESS_MARKERS` fails `test_does_not_match_processes_that_merely_mention_maya`.

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,99 @@
...
+    for proc, cmdline in orphans:
+        print(f"[license-cleanup]   pid={proc.pid} {cmdline[:160]}")
+        try:
+            proc.terminate()
+        except (psutil.NoSuchProcess, psutil.AccessDenied) as exc:
+            print(f"[license-cleanup]   pid={proc.pid} terminate failed: {exc}")
+
+    _, still_alive = psutil.wait_procs([proc for proc, _ in orphans], timeout=15)
+    for proc in still_alive:
+        print(f"[license-cleanup]   pid={proc.pid} ignored SIGTERM; sending SIGKILL")
```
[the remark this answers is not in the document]
andrew-fenton: Log wording fixed in 748561a: now `did not exit within 15s; killing`, with a comment noting the escalation only fires on POSIX. Confirmed from the installed psutil, whose `terminate()` docstring reads "On Windows this is an alias for kill()", and `_pswindows.py` maps `SIGTERM` straight to `proc_kill`.

On the license-timing concern: reclaim was immediate in practice. The 2026-09-10 run terminated 22 orphans, the endpoint dropped to zero sessions, and the suite checked out licenses successfully minutes later in the same build. No dependence on a server reclaim interval.

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,98 @@
...
+        return
+
+    try:
+        import psutil
+    except ImportError:
+        _log(f"{phase}: psutil unavailable; skipping orphan cleanup")
+        return
+
+    # Guards against self-termination if this is ever called from mayapy.
+    protected_pids = {os.getpid()}
```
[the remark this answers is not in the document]
andrew-fenton: Checked this, and concurrent builds cannot occur on this fleet, so not changing it.

Phase timings show the contention is handled by queueing, not co-tenancy:

```
exec start           end        status     queued
2026-09-03T15:25:02  15:34:21   SUCCEEDED       0s
2026-09-03T15:34:37  16:07:35   STOPPED       131s
```

The second build waited 131s in `QUEUED` and started 16s after the first finished. Across 22 builds over 8 days there are zero overlapping execution windows. `concurrentBuildLimit` is `null`, so serialization comes from the fleet `baseCapacity: 1`. Raising capacity would add instances, not co-tenants, each with its own process table.

Two further notes. The lock files in `setup-runner.py` look like dead code: the buildspec `pre_build` already runs `find /tmp -name "codebuild-*" ! -path "$CURRENT_DIR" | xargs rm -rf`, which would be deleting the working directory of a live build if hosts were shared. And the suggested guard would not fix the case described, since skipping processes newer than the current build fails when the sibling started first, which is the likely ordering.

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,99 @@
...
+    CI runs on CodeBuild reserved-capacity fleets whose hosts are reused between
+    builds. When a Maya process dies without releasing its Autodesk license --
+    for example when ``maya.standalone.initialize()`` raises and the test
+    fixture's ``uninitialize()`` teardown never runs -- the checkout is never
+    returned, and the orphan keeps its license session alive indefinitely. Once
+    enough accumulate, later checkouts are refused and Maya reports it as the
+    misleading "Error encountered when initializing Maya - Please check for
+    sufficient disk space and necessary write permissions of MAYA_APP_DIR."
+
+    No Maya process should be running before the suite starts, so anything
```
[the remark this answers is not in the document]
andrew-fenton: Same as the thread on line 69: concurrent builds on one host do not occur here. Phase timings show queueing rather than co-execution, with zero overlapping execution windows across 22 builds.

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,99 @@
...
+    """
+    orphans = []
+    for proc in psutil.process_iter(["pid", "cmdline"]):
+        if proc.pid in protected_pids:
+            continue
+        try:
+            cmdline = " ".join(proc.info.get("cmdline") or [])
+        except (psutil.NoSuchProcess, psutil.AccessDenied):
+            continue
+        if any(marker in cmdline for marker in _MAYA_PROCESS_MARKERS):
```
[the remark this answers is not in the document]
andrew-fenton: Leaving as substring matching for now.

The main motivation was the concurrency scenario, which does not occur on this fleet (see the line 69 thread). Against the change: on Linux `mayapy` is a shell wrapper, so name matching would stop matching it, which is a behaviour change to code already validated on a real host.

The `"Render"` example is the strongest argument and is now covered by a test. Job parameters in this suite contain `rs_<RenderLayer>_<Camera>`, so adding that marker fails `test_does_not_match_processes_that_merely_mention_maya`. Happy to revisit if the wrapper behaviour is not wanted.

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,98 @@
...
+    orphans = find_orphaned_maya_processes(psutil, protected_pids)
+    if not orphans:
+        _log(f"{phase}: no orphaned Maya processes found")
+        return
+
+    _log(f"{phase}: terminating {len(orphans)} orphaned Maya process(es):")
+    for proc, cmdline in orphans:
+        _log(f"  pid={proc.pid} {cmdline[:160]}")
+        try:
+            proc.terminate()
```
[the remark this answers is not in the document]
andrew-fenton: Duplicate of the line 87 thread; see the reply there. Log wording fixed in 748561a.

test/integ/test_maya_submitters.py
```diff
@@ -32,10 +32,13 @@ def initialize_maya():
     )
 
     qt_application = QtWidgets.QApplication(sys.argv)
-    yield show_maya_render_submitter, on_create_job_bundle_callback
-
-    qt_application.shutdown()
-    maya.standalone.uninitialize()
+    try:
```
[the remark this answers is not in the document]
andrew-fenton: Fixed: the `try` now opens immediately after `initialize()`.

`initialize()` stays outside it. If it fails no license was acquired, and an `uninitialize()` error from the `finally` would mask the original error.

The second point is not correct: pytest resumes a yield-fixture with `next()`, not `throw()`, so a test failure never reaches the generator and `shutdown()` is not skipped. Verified with a failing test against a yield fixture.

pipeline/run-integ-tests.py
```diff
@@ -44,11 +144,14 @@ def main():
...
+    try:
+        result = subprocess.run(
             ["mayapy", "-m", "pytest", "--no-cov", "test/integ", "-vvv", "--numprocesses=1"]
-        ).returncode
-    )
+        )
+    finally:
+        _cleanup("post-test")
+
+    sys.exit(result.returncode)
```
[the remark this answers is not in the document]
andrew-fenton: This is a known trade-off I made by going with the pre and post-cleanup approach. Any orphaned fleets that cannot be cleaned up post-run will be cleaned up at the start of the next run.

 Modified docs to reflect this tradeoff.

test/integ/test_maya_submitters.py
```diff
@@ -22,20 +22,25 @@ def initialize_maya():
     """
     Fixture that ensures Maya is open and close after the test runs.
     """
+    # Everything after initialize() runs with a license checked out, so it all
+    # belongs inside the try. initialize() itself does not: if it fails there is
+    # no license to release, and an uninitialize() error would mask the original
```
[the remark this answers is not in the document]
andrew-fenton: The runner docstring was the stale one; it carried the initial theory. Every orphan observed was an adaptor subprocess that outlived a cancelled or timed-out run, never the pytest host. Both files now say that (6d5e07d).

Not taking the code change: the case is unobserved, and a session-scoped autouse fixture failure errors every dependent test, so there are no remaining tests to starve.

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,123 @@
...
+
+    _log(f"{phase}: terminating {len(orphans)} orphaned Maya process(es):")
+    for proc, description in orphans:
+        _log(f"  pid={proc.pid} {description[:160]}")
+        try:
+            proc.terminate()
+        except (psutil.NoSuchProcess, psutil.AccessDenied) as exc:
+            _log(f"  pid={proc.pid} terminate failed: {exc}")
+
+    _, still_alive = psutil.wait_procs([proc for proc, _ in orphans], timeout=15)
```
TanoVip123, line 99: Why do we call terminate() here if and wait 15 s if we are calling kill() anyway? From the comment, it looks like that terminate is already kill() on Windows so maybe we can just call kills from the get going?

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,123 @@
...
+            continue
+        cmdline_parts = proc.info.get("cmdline")
+        if cmdline_parts is None:
+            name = proc.info.get("name") or ""
+            if name.lower() in _MAYA_PROCESS_NAMES:
+                orphans.append((proc, f"name={name} (command line unreadable)"))
+            continue
+        cmdline = " ".join(cmdline_parts)
+        if any(marker in cmdline for marker in _MAYA_PROCESS_MARKERS):
+            orphans.append((proc, cmdline))
```
TanoVip123, line 55: nit: do you think this is cleaner instead of continue?
```
else:
    cmdline = " ".join(cmdline_parts)
    if any(marker in cmdline for marker in _MAYA_PROCESS_MARKERS):
        orphans.append((proc, cmdline))
```

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,123 @@
...
+environment variables so the adaptor's subprocess can find mayapy and renderer
+plugins.
 """
 
 import os
 import platform
 import subprocess
 import sys
 
+_MAYA_PROCESS_MARKERS = ("mayapy", "maya_client.py", "maya.bin", "maya.exe")
```
TanoVip123, line 14: What is the difference between _MAYA_PROCESS_NAME and _MAYA_PROCESS_MARKERS? and why is one a frozenset and the other a set?

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,123 @@
...
+# can hold a license. Compared lowercased, exact match.
+_MAYA_PROCESS_NAMES = frozenset(
+    {"mayapy", "mayapy.bin", "mayapy.exe", "maya", "maya.bin", "maya.exe"}
+)
+
+
+def _log(message):
+    """Log a cleanup message.
+
+    flush=True keeps ordering against the pytest subprocess, which writes to the
```
jairaws, line 26: nit: `flush` is not part of the interface. This comment makes more sense as an inline comment, than as a function level one.

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,123 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jairaws, line 1: Could we type variables when declared  and in the functions signature?

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,123 @@
...
 """
 
 import os
 import platform
 import subprocess
 import sys
 
+_MAYA_PROCESS_MARKERS = ("mayapy", "maya_client.py", "maya.bin", "maya.exe")
+
+# Fallback for processes whose command line cannot be read: executable names that
```
jairaws, line 16: Q: In what scenario can't the command line be read?

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,123 @@
...
+# can hold a license. Compared lowercased, exact match.
+_MAYA_PROCESS_NAMES = frozenset(
+    {"mayapy", "mayapy.bin", "mayapy.exe", "maya", "maya.bin", "maya.exe"}
```
jairaws, line 19: Q: One of the names simply `maya`. Should we just be checking if the processes name (line) contains maya?

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,123 @@
...
+    killed by timeout or cancellation dies on SIGTERM without unwinding, so its
+    orphans are reclaimed by the next build's pre-test sweep. CodeBuild only, so
+    it cannot kill a Maya a developer opened.
+    """
+    if not os.environ.get("CODEBUILD_BUILD_ID"):
+        _log(f"{phase}: not running in CodeBuild; skipping orphan cleanup")
+        return
+
+    try:
+        import psutil
```
jairaws, line 78: Q: I wonder why we are importing psutil within this function logic? And whether we had issues importing it before?

pipeline/run-integ-tests.py
```diff
@@ -1,17 +1,123 @@
...
+    except ImportError:
+        _log(f"{phase}: psutil unavailable; skipping orphan cleanup")
+        return
+
+    # Guards against self-termination if this is ever called from mayapy.
+    protected_pids = {os.getpid()}
+
+    orphans = find_orphaned_maya_processes(psutil, protected_pids)
```
jairaws, line 86: Q: Why do we need to pass psutil as a parameter? It doesn't seem to be holding any state.

pipeline/run-integ-tests.py
andrew-fenton: Yes, will do.

pipeline/run-integ-tests.py
andrew-fenton: True, fixed.

pipeline/run-integ-tests.py
andrew-fenton: Chose to inline the import because I intended the cleanup mechanism to be best-effort. I think it's better for a valid integration test run to succeed even if the cleanup fails as it is not an issue stemming from the tests or intended changes themselves.

As for your second question: no issues importing before, simply defensive programming.

pipeline/run-integ-tests.py
andrew-fenton: This is related to the inline import of `psutil` to preserve best-effort cleanup. When imported inline, the module won't be in scope if not passed as a parameter and I think passing the parameter is better than importing again.

pipeline/run-integ-tests.py
andrew-fenton: We technically could just call `kill()` for Linux too but I prefer terminating processes gracefully and escalate to kill if needed. Reason is killing processes forcefully can leave child processes alive and we cannot guarantee that these child processes will be caught by our initial scan using our markers.

pipeline/run-integ-tests.py
andrew-fenton: `_MAYA_PROCESS_NAMES` was a fallback for when `psutil` reported a process's command line as `None`. We could match on the process name instead.

I removed it because a process name is not necessarily an accurate representation of what the process is doing: anything running under a generic interpreter reports python3 or bash, so the identifying detail sits in the command line rather than the process name.

pipeline/run-integ-tests.py
andrew-fenton: `psutil` reports the command line as None when either `AccessDenied` or `ZombieProcess` is raised.

I removed it since these processes are not really relevant:
- `AccessDenied`: we cannot confidently identify the process so we should not terminate
- `ZombieProcess`: process has already exited, not holding any license

Also, mentioned in a comment before but it's more robust to match on command line than name for this sweeper.

pipeline/run-integ-tests.py
andrew-fenton: The `"maya"` string check was an exact match on the process name, so it only matched a binary named exactly maya. It is now removed for reasons laid out in other comments, but the strict check was intentional.

The command-line markers are matched by substring, so an only `"maya"` check would match any command line containing `"maya"`, which is not always a Maya process that we want to terminate (e.g. tailing a log file with maya in its name).

jairaws (approved):

TanoVip123 (approved):

